PR #2, opened 2010-10-03
Reltime

pcc:
This change would fix pluralisation issues (e.g. "1 minutes ago") across all languages but unfortunately would mean we end up losing the translation for a few languages (looking at the current Froyo frameworks/base repository we lose ca, gl, hu, id, sk but gain el, pt, tr). An alternative fix could be to fix the pluralisation issue in connectbot (code duplication), or possibly port our translations to frameworks/base.  Let me know what you think.


kruton:
Actually, I think this is prudent now.

PR #9, opened 2011-03-20
retry connection patch

shaiay:
This patch adds several retries to connect before giving up. This came up because when switching from wifi to 3G, it takes some time for the 3G network to come up, and the first reconnect attempt fails with the "host unreachable" error. With this patch connectbot becomes quite robust with regards to connections.

I'm not sure if this patch has any undesirable side effects, but in such a case I would be happy to add it as a configuration option to be enabled only by those who need it.

And lastly, this is my first time writing java, and only after looking at the SDK for a few days, so I apologize if I made any glaring mistakes, and would be happy to hear any criticism and imporve.

Oh, and thanks for connectbot!

Shai

### DIFF
--- a/src/org/connectbot/service/ConnectivityReceiver.java
+++ b/src/org/connectbot/service/ConnectivityReceiver.java
@@ -65,22 +65,20 @@ public class ConnectivityReceiver extends BroadcastReceiver {
             Log.w(TAG, "onReceived() called: " + intent);
             return;
 		}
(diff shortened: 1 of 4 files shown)


adarazs:
I understand this is a 5 year old pull request, however the issue still exists on 1.8.6 (latest version available on the Play Store) and this pull request might solve it. Could somebody pick this up and merge it? It would be really useful.

PR #13, opened 2011-06-13
Allow other Android Apps to use ConnectBot as an ssh-agent

rtyley:
I've recently released Agit, a Git client for Android (https://market.android.com/details?id=com.madgag.agit). In order to support git+ssh functionality, I created this patch of ConnectBot which allows other android apps to use it in the same manner as an ssh-agent. This means that Agit doesn't need to actually manage or even store SSH keys itself, which is pretty cool - as ConnectBot already has the UI & infrastructure to manage SSH keys, I don't want to have to do it again...!

The functionality is exposed using AIDL, defining a small two-method interface, getIdentities() & sign() :

https://github.com/rtyley/madgag-ssh/blob/master/ssh-android/src/main/java/com/madgag/ssh/android/authagent/AndroidAuthAgent.aidl

The patch adds the AndroidAuthAgent.java generated from this AIDL definition to ConnectBot, and exposes it using a AuthAgentService to external Apps with the `org.openintents.ssh.permission.ACCESS_SSH_AGENT` permission so that they can bind to the agent and use it to provide public-keys and sign the ssh-challenges that they encounter.

Just to go over that again, it's the external app (e.g. Agit) that creates the network connection to the remote ssh server, using it's _own_ ssh libraries, it's just that it binds to ConnectBot in order to get a list of available identities (ie public keys), and then later, when the remote ssh server issues a challenge, Agit will call the auth-agent again to sign the issued ssh cryptographic-challenge.

The bit of Agit's code where it does this is here:

https://github.com/rtyley/agit/blob/agit-parent-1.7/agit/src/main/java/com/madgag/agit/ssh/AndroidSshSessionFactory.java

It works really well - it's a symphony of cryptography and inter-process communication!

Hope you like it,
Roberto


joonas-fi:
Why's this cool feature just collecting dust? :)

pyhedgehog:
Any hope for progress on this?

nwf:
What's the status of this?  I'd love to have this in mainline as well as the forks so that other programs (e.g. VLC's SFTP suppot) could make use of it.

lpalgarvio:
getting closer to 6 years...

fengshaun:
Any love for this? I'm looking to integrate this functionality into syncopoli.

JJRcop:
The status of affairs appears to be that @kruton gave feedback requesting changes such that there is user confirmation, then @rtyley opened a discussion, starting that needing to request confirmation each time would be annoying for his agit app, which this patch was originally made for. Since then there has been no response from either party.

If I may add to the discussion, I think being able to trust an app using a checkbox on the confirmation screen would resolve the concerns of both parties, this could simply write the name or id of the app to a "trusted_apps_ssh-agent" variable or file.

PR #23, opened 2013-06-19
Added a "close on disconnect" option 

mendhak:
Added a close-on-disconnect option.  This enables/facilitates automation (for example: you can invoke a connectbot shortcut from tasker and this will ensure that the screen isn't stuck on the disconnect prompt).

Close-on-disconnect will not take effect if stay-connected is checked.

screenshot_2013-06-19-07-02-29

Slightly modified from [peffreyking's branch](https://code.google.com/r/peffreyking-connectbot/source/detail?r=6db98fe935581f27fd86a50748497a45ce53b9f3&name=quick-disconnect).


arne182:
Can this be added in the next release please

PR #25, opened 2013-10-09
Fixnetwork

giuliolunati:
Two commits for fixing Issue 417 (http://code.google.com/p/connectbot/issues/detail?id=417) and another similar.


kruton:
Do you have a traceback for the "Fix crash when network go on."? What is causing the exception?

giuliolunati:
I switched to [http://code.google.com/p/connectbot-improved/](connecbot-improved)

kruton:
Uh, okay.

PR #28, opened 2014-01-12
Add support for closing the connection automatically on disconnect

wmoss:
This setting is specifically useful for me (see below) and I understand if you don't want to merge it back but I figured I would send the PR incase you did (since it is just an extra setting). Also, I don't have translations for any of the other languages (nor a good way to acquire them). 

**Full story:** I use connectbot to connect to a beagleboard that opens my garage and front door. I have separate users on the beagleboard that run appropriate commands (i.e. open garage) and then exit (dropping the connection). I keep widgets on my homescreen for these tasks/users and it's annoying that I have to click "Yes" to close the connect after clicking one of the widgets. This setting solves that.


kruton:
I merged #23 last week, so I'll close this.

PR #35, opened 2014-09-04
Change text and some translations of hostpref_nickname_title to "name".

htgoebel:
I changed them to (the translation of) just "name" (not "nickname"), as this is what Android uses in VPN settings.

For German and Frensh the translations have been completly wrong (not translating "nickname") and missleading.

### DIFF
--- a/res/values-da/strings.xml
+++ b/res/values-da/strings.xml
@@ -107,7 +107,7 @@
   <string name="list_update_daily">Dagligt</string>
   <string name="list_update_weekly">Ugenligt</string>
   <string name="list_update_never">Aldrig</string>
-  <string name="hostpref_nickname_title">Kaldenavn</string>
+  <string name="hostpref_nickname_title">Navn</string>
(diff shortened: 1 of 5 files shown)


htgoebel:
"Pseudonyme" is even more inapplicable. Just use "Name".

htgoebel:
Addendum:

* For German "Spitzname" is wrong, nobody would use this. Just use "Name".
* For Franch: I'm french-speaking, but "Pseudonyme" is not applicable at all. According to http://www.finedictionary.com/pseudonym.html: "pseudonym a fictitious name used when the person performs a particular social role". So either use "Sobriquet" (which is also a word in English, "Surnom" (which us more like the family name) or just "Nom".

kruton:
I've put your suggestions into the [translation website](https://translations.launchpad.net/connectbot/trunk/+pots/fortune) but it will take a while to change "Nickname:" to "Name:"

PR #68, opened 2015-03-28
Add Serbian translation

pejakm:


kruton:
Thanks for the translations!

PR #91, opened 2015-07-13
Switch the quick connect entry in the hostlist to a LinearLayout.

jklein24:
This fixes a bug in landscape where the input is not visible at all.
See
http://stackoverflow.com/questions/23982098/android-relativelayout-layout-above-appears-to-behave-incorrectly-with-layout-a.


kruton:
Does this fix issue 86?

jklein24:
Yep, I believe it does.

PR #95, opened 2015-07-14
Add a tab button to the emulated keys above the soft keyboard.

### LINKED ISSUE #92: Add a tab button where ctrl and esc are
The default android virtual keyboard has no tab button, so this would be convenient.

jklein24:
Fixes #92


jklein24:
@kruton let me know if there's anything I should do wrt compression on the png or translations for the new string.

kruton:
Could you attach a screenshot of what it looks like after this change?

kruton:
Also I have a project that uses gimp to automatically create assets based on a scheme program. It might be worth converting the soft buttons to this or making them into TextView instead.

Take a look at https://github.com/connectbot/assets

jklein24:
tab_button


I'll check out the asset creator too.

jklein24:
tab_button

jklein24:
fwiw, +1 to making these buttons into actual text buttons.

Issue #93, opened 2015-07-14
Add a disconnect button to the active connection notification


jklein24:

kruton:
Closed by #103.

PR #99, opened 2015-07-15
Add the v4 support library and update the notification API.

jklein24:
Switch to using NotificationCompat.Builder. This will allow for
the creation of a "disconnect all" button in the notification
(see #93).

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -4,13 +4,13 @@ apply from: '../config/quality.gradle'
 apply from: '../config/translations.gradle'
 
 android {
-    compileSdkVersion 20
-    buildToolsVersion "20.0.0"
+    compileSdkVersion 22
+    buildToolsVersion "21.0.0"
 
     defaultConfig {
         applicationId "org.connectbot"
         minSdkVersion 4
-        targetSdkVersion 15
+        targetSdkVersion 22
         compileOptions {
             sourceCompatibility JavaVersion.VERSION_1_5
             targetCompatibility JavaVersion.VERSION_1_5
@@ -32,6 +32,10 @@ android {
 
     signingConfigs {
         release
+    }
+
+    dependencies {
+        compile "com.android.support:support-v4:22.0.+"
     }
 
     buildTypes {
--- a/app/src/main/java/org/connectbot/service/ConnectionNotifier.java
+++ b/app/src/main/java/org/connectbot/service/ConnectionNotifier.java
@@ -44,7 +44,6 @@ import android.support.v4.app.NotificationCompat;
  */
 public abstract class ConnectionNotifier {
 	private static final int ONLINE_NOTIFICATION = 1;
-	private static final int ONLINE_NOTIFICATION_DISCONNECT = 1;
 	private static final int ACTIVITY_NOTIFICATION = 2;
 
 	public static ConnectionNotifier getInstance() {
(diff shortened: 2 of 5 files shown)


jklein24:
Looks like the build bot is failing because it can't find the support libs. How can I make sure they're installed on the bot?

app/src/main/java/org/connectbot/service/ConnectionNotifier.java
```diff
@@ -56,16 +58,17 @@ protected NotificationManager getNotificationManager(Context context) {
...
+				.setSmallIcon(R.drawable.notification_icon)
+				.setWhen(System.currentTimeMillis());
 
-		return notification;
+		return builder;
 	}
 
 	protected Notification newActivityNotification(Context context, HostBean host) {
-		Notification notification = newNotification(context);
+		NotificationCompat.Builder notification = newNotificationBuilder(context);
```
kruton: Might as well rename the variable to "builder" to be more clear.

app/src/main/java/org/connectbot/service/ConnectionNotifier.java
```diff
@@ -56,16 +58,17 @@ protected NotificationManager getNotificationManager(Context context) {
 		return (NotificationManager) context.getSystemService(Context.NOTIFICATION_SERVICE);
 	}
 
-	protected Notification newNotification(Context context) {
-		Notification notification = new Notification();
-		notification.icon = R.drawable.notification_icon;
-		notification.when = System.currentTimeMillis();
+	protected NotificationCompat.Builder  newNotificationBuilder(Context context) {
```
kruton: There are two spaces in here instead of one.

app/src/main/java/org/connectbot/service/ConnectionNotifier.java
```diff
@@ -43,6 +44,7 @@
  */
 public abstract class ConnectionNotifier {
 	private static final int ONLINE_NOTIFICATION = 1;
+	private static final int ONLINE_NOTIFICATION_DISCONNECT = 1;
```
kruton: Does this work? Not sure what the constant values mean here.

app/build.gradle
```diff
@@ -4,13 +4,13 @@ apply from: '../config/quality.gradle'
...
-    compileSdkVersion 20
-    buildToolsVersion "20.0.0"
+    compileSdkVersion 22
+    buildToolsVersion "21.0.0"
 
     defaultConfig {
         applicationId "org.connectbot"
         minSdkVersion 4
-        targetSdkVersion 15
+        targetSdkVersion 22
```
kruton: Do we need to update the targetSdkVersion for the support library?

kruton:
Maybe modify the `.travis.yml` and add `extra-android-m2repository` under the android section.

app/build.gradle
jklein24: It's not totally required, but I was getting warnings to use the new one.

app/src/main/java/org/connectbot/service/ConnectionNotifier.java
jklein24: Ah sorry, this snuck in there when I was playing with adding the disconnect button. Removed.

app/src/main/java/org/connectbot/service/ConnectionNotifier.java
jklein24: Done.

app/src/main/java/org/connectbot/service/ConnectionNotifier.java
jklein24: Done.

jklein24:
travis seems happy now

PR #103, opened 2015-07-22
Add a disconnect button to the active connection notification

jklein24:


app/src/main/java/org/connectbot/HostListActivity.java
```diff
@@ -394,13 +431,37 @@ public void onClick(DialogInterface dialog, int which) {
...
+		// Copy the bridges list because bridges are removed from the array when disconnected.
+		for (TerminalBridge bridge : new ArrayList<TerminalBridge>(bound.bridges)) {
+			bridge.dispatchDisconnect(true);
+		}
+		updateHandler.sendEmptyMessage(-1);
+		waitingForDisconnectAll = false;
+
+		if (closeOnDisconnectAll) {
+			// Clear the intent so that the activity can be relaunched without closing.
+			// TODO(jlklein): Find a better way to do this.
```
jklein24: @kruton, do you know of a better way to do this? The details of figuring out when this activity should actually be closed are really tricky and seem error-prone. Basically, I only want to close the activity if it was only launched to do the disconnect.

app/src/main/java/org/connectbot/HostListActivity.java
```diff
@@ -129,11 +145,33 @@ public void onStop() {
...
 	}
 
 	@Override
 	public void onResume() {
 		super.onResume();
+
+		// Must disconnectAll before setting closeOnDisconnectAll to know whether to keep the
+		// activity open after disconnecting.
+		if ((getIntent().getFlags() & Intent.FLAG_ACTIVITY_LAUNCHED_FROM_HISTORY) == 0 &&
+				getIntent().getAction() == DISCONNECT_ACTION) {
```
kruton: DISCONNECT_ACTION.equals(getIntent().getAction())

app/src/main/java/org/connectbot/HostListActivity.java
kruton: This should probably go into `TerminalManager`. Also, if you receive the `DISCONNECT_ACTION`, you should probably always `finish()`. The user doesn't want to continue interacting with the app if they're closing everything, right?

app/src/main/java/org/connectbot/HostListActivity.java
jklein24: Done.

app/src/main/java/org/connectbot/HostListActivity.java
jklein24: Perhaps, but then should that also be the case if the user disconnects individually from all connections using the context menus? I feel like if the user was already in the hostlistactivity and then presses disconnect all, they should stay where they were. I could be convinced either way, though. 

sigh... totally missed that there's already a `disconnectAll` in TerminalManager. Updated.

PR #107, opened 2015-07-25
Unit test

kruton:
@jklein24 this is an example unit test.


kruton:
Whoops, forgot to include the bridges -> getBridges refactor.

PR #112, opened 2015-07-29
Mouse actions: Start copying on drag and paste on wheel click.

jklein24:


kruton:
Does this break the swipe to next screen stuff? This is a better use of primary click at any rate.

Maybe we can have mouse pointers hovering on the right or left edge of the screen flip to the next terminal?

jklein24:
Hmm I suppose this breaks the swipe between screens when using a mouse, but it still works with your finger. I find swipe a bit awkward for this purpose with a mouse anyway, particularly on large screens. I don't think it's too big of a deal. There's also a pending request from @bengolds to add tabs on large screens anyway.

jklein24:
uh oh, my git history went nuts... fixing it.

jklein24:
Fixed.

jklein24:
Strange stuff going on here... it seems like connectbot:master is missing all the merged commits from my previous PRs somehow. Can you take a look at what happened on master?

kruton:
Looks like my bad. I merged in 41a1eb43266d64c3fe8f519ed00d16f6436d1cae again to fix the history.

jklein24:
OK, I'm fairly convinced everything is sane now.

PR #119, opened 2015-08-05
ConsoleActivity: Switch from ViewFlipper to ViewPager.

jklein24:
This simplifies the code for creating terminals, removes lots of code,
and totally eliminates the need for swipe handling.
From the UI side, the swipe now sticks to the user's finger and looks
much nicer. It will also make adding tabs much simpler.

### DIFF
--- a/app/src/main/java/org/connectbot/ConsoleActivity.java
+++ b/app/src/main/java/org/connectbot/ConsoleActivity.java
@@ -173,6 +173,7 @@ public class ConsoleActivity extends Activity {
 		}
 
 		public void onServiceDisconnected(ComponentName className) {
+			adapter.notifyDataSetChanged();
 			updateEmptyVisible();
 			bound = null;
 		}
@@ -195,6 +196,7 @@ public class ConsoleActivity extends Activity {
 			// they are sending nickname and requested
 			TerminalBridge bridge = (TerminalBridge) msg.obj;
 
+			adapter.notifyDataSetChanged();
 			if (bridge.isAwaitingClose())
 				closeBridge(bridge);
 		}
@@ -205,7 +207,6 @@ public class ConsoleActivity extends Activity {
 	 */
 	private void closeBridge(final TerminalBridge bridge) {
 		synchronized (pager) {
-			adapter.notifyDataSetChanged();
 			updateEmptyVisible();
 			updatePromptVisible();
 
--- a/app/src/main/java/org/connectbot/ConsoleActivity.java
+++ b/app/src/main/java/org/connectbot/ConsoleActivity.java
@@ -169,7 +169,7 @@ public class ConsoleActivity extends Activity {
 			adapter.notifyDataSetChanged();
 			requestedIndex = bound.getBridges().indexOf(requestedBridge);
 
-			pager.setCurrentItem(requestedIndex == -1 ? 0 : requestedIndex);
+			setDisplayedTerminal(requestedIndex == -1 ? 0 : requestedIndex);
 		}
 
 		public void onServiceDisconnected(ComponentName className) {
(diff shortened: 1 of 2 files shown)


adb2001:
Nice work !

kruton:
This doesn't update the input box in the correct window. I couldn't repro this, but it happened once:

1. Open local term
2. Go back to host list
3. Open SSH term
4. Page back and forth between the terminals
5. "Password:" dialog shows up on local term

Here's an easier one:

1. Open local term
2. Open SSH term
3. Wait for "Password:" dialog to show up
4. Go to menu and "Disconnect"
5. "Password:" dialog shows up on local term

kruton:
Oh, it might be that the terminal is almost slid all the way over to the first window when the password prompt pops up.

app/src/main/java/org/connectbot/ConsoleActivity.java
```diff
@@ -169,20 +166,13 @@ public void onServiceConnected(ComponentName className, IBinder service) {
...
-
-				// check to see if this bridge was requested
-				if (bridge == requestedBridge)
-					requestedIndex = currentIndex;
-			}
+			adapter.notifyDataSetChanged();
+			requestedIndex = bound.getBridges().indexOf(requestedBridge);
 
-			setDisplayedTerminal(requestedIndex);
+			pager.setCurrentItem(requestedIndex == -1 ? 0 : requestedIndex);
```
kruton: I think this is missing an `updatePromptVisible`, etc. Perhaps you should wrap all that in a new `setDisplayedTerminal` that calls `pager.setCurrentItem(…)` in addition to `updatePromptVisible()` and `updateEmptyVisible()`.

app/src/main/java/org/connectbot/ConsoleActivity.java
```diff
@@ -985,15 +938,16 @@ protected void onNewIntent(Intent intent) {
...
 			} else {
-				final int flipIndex = getFlipIndex(requestedBridge);
+				final int flipIndex = bound.getBridges().indexOf(requestedBridge);
 				if (flipIndex > requestedIndex) {
 					requestedIndex = flipIndex;
 				}
 			}
 
-			setDisplayedTerminal(requestedIndex);
+			pager.setCurrentItem(requestedIndex);
```
kruton: I think this is also missing an `updatePromptVisible()`, etc., as above and should be wrapped in a call to `setDisplayedTerminal(…)`.

jklein24:
OK I added the setDisplayedTerminal function back with the appropriate update calls, but am still able to repro this. The way I can repro it is if I open the ssh term and then swipe back to local before the prompt shows up. Looking into it more right now.

jklein24:
OK, tracked it down. Turns out that the issue was in getCurrentTerminalView. It turns out the child index in the ViewPager != to the position according to the adapter. This means doing `pager.getChildAt(pager.getCurrentItem())` returns the wrong view sometimes. Tagging the views with their positions as they're inserted fixes the issue:

http://stackoverflow.com/questions/12854783/android-viewpager-get-the-current-view

jklein24:
@adb2001, I'm working on adding tabs in https://github.com/jklein24/connectbot/tree/tabs, but probably only for large screens because they take up quite a bit of real estate.

kruton:
No, this doesn't work either it seems. If you start a local terminal (gets tag `0`) then start an SSH session (gets tag `1`), wait for the prompt to come up, swipe back to the local session, select disconnect from the menu, the `getCurrentTerminalView()` will return `null` because the `pager.getCurrentItem()` returns `0` which ends up being null because the local term was tagged with `0`

kruton:
Okay, I changed it to use `pager.getCurrentItem()` to query the `adapter` for what item is at that position. That works.

PR #121, opened 2015-08-08
Reset auto hide timer when an arrow button is clicked to allow multiple click.

alescdb:
- Reset auto hide timer when an arrow button is clicked to allow multiple arrow click. 
- Create onEmulatedKeyClicked() via a global OnClickListener (emulatedKeysListener) handling all button events (except keyboard) since they all share the same code, we could also implement OnClickListener directly in ConsoleActivity to avoid the use of a global variable.
- Remove unused keyboard png (esc, tab, ctrl)

PS: I'm not really comfortable with github yet, please excuse me if I'm doing something wrong here by re-submitting a new pull request.

### DIFF
--- a/app/src/main/java/org/connectbot/ConsoleActivity.java
+++ b/app/src/main/java/org/connectbot/ConsoleActivity.java
@@ -18,6 +18,7 @@
 package org.connectbot;
 
 import java.lang.ref.WeakReference;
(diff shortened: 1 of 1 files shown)


kruton:
Looks like this accidentally reverts the ViewPager changes.

PR #122, opened 2015-08-09
Reset autohide timer when an arrow button is clicked to allow multiple click

alescdb:
- Reset auto hide timer when an arrow button is clicked to allow multiple arrow click.
- Create onEmulatedKeyClicked() via a global OnClickListener (emulatedKeysListener) handling all button events (except keyboard) since they all share the same code, we could also implement OnClickListener directly in ConsoleActivity to avoid the use of a global variable.
- Remove unused keyboard png (esc, tab, ctrl)
- Add haptic feedback to all keys except tab and keyboard.

PS: This time the patch looks good.


kruton:
Any reason why the tab doesn't vibrate? Are you trying to match existing behavior?

alescdb:
The main reason is that Tab always produce a visible result, and when it does not (eg in "bash" for auto completion when there is more than one match) it produce a bell.

That was the idea, now if you prefer I can remove the Tab exception :)

kruton:
Yeah, let's remove that because the delay might be so great on the SSH link over mobile networks. Otherwise this looks great. Thanks.

alescdb:
Good point, Changes done.

PR #126, opened 2015-08-11
Copy editor

alescdb:
I found the default "Copy" very painful... on small screen with a pretty small font and big fingers, I almost always miss the line or the part of text I want to copy and ends up doing my "Copy" over and over until I get it right. Also, the console has no margins, so copying from the first character of a line is mission impossible (almost).

The idea in this pull request is to show the scrollback buffer of the current console to the user in an EditText (in a new Activity) with large margins, allowing to use the default Android copy/paste system.

I haven't found a way to restore CR/LF to their proper places, it seems it's impossible without adding a buffer somewhere (I think TerminalView::mAccessibilityBuffer does somthing like that).

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -68,6 +68,7 @@
 		<activity android:name=".HelpActivity" android:configChanges="keyboardHidden|orientation" />
 		<activity android:name=".HelpTopicActivity" android:configChanges="keyboardHidden|orientation" />
 		<activity android:name=".ColorsActivity" android:configChanges="keyboardHidden|orientation" />
+		<activity android:name=".CopyEditActivity" android:configChanges="keyboardHidden|orientation" />
 
 		<service android:name="org.connectbot.service.TerminalManager"
 			android:configChanges="keyboardHidden|orientation"
--- a/app/src/main/java/org/connectbot/ConsoleActivity.java
+++ b/app/src/main/java/org/connectbot/ConsoleActivity.java
@@ -17,18 +17,6 @@
 
 package org.connectbot;
 
-import java.lang.ref.WeakReference;
-import java.util.ArrayList;
-import java.util.List;
-
-import org.connectbot.bean.HostBean;
-import org.connectbot.bean.SelectionArea;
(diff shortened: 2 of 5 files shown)


kruton:
I agree this is a problem, but I kind of wanted a better copy & paste experience all around. I'd rather have a "loupe view" when touch is being used on a phone. Also there are a couple use cases with different devices:

1. The loupe view for touch on the phone.
2. Using selection with a mouse:
  * On a TV where the remote control has a touchpad or joystick.
  * Tablet with a bluetooth mouse where there is a cursor for selection.

kruton:
(Also tagging @jklein24 who has discussed this with me before.)

alescdb:
Won't it be "more simple" to extend or mimic the behavior of TextView in TerminalView and use the default Android copy/paste system ?

jklein24:
This seems very similar to what Juice SSH does right now for copy. I definitely agree that, for the reasons you've mentioned, this is a nicer solution for touch selection (there's a reason all of Android text selection works this way).

However, I'm not sure that I like the idea of having "copy" and "copy editor" menu items separately. As a user, I'd want the app to choose the mode that makes the most sense in the right context. As @kruton noted, if I'm using a bluetooth mouse, copying should just work like on a desktop. I just feel like having multiple copy modes in the menu is a bit confusing as a user.

I also don't totally love the idea of opening a new activity to do copying. It felt really weird to me the first time I saw this in JuiceSSH. My ideal here would be that we make it feel like the same view in a copy mode like an ActionMode. I wonder if we could silently swap out the terminalview for an EditText when the user enters copy mode to get all the Android text selection goodness. It's hacky, but could result in the nicest UX.

jklein24:
cc @bengolds

alescdb:
I've deleted the branch by mistake !
Whatever, we all agreed that it was not a good solution.

For testing purpose I've made :

In TerminalView :
- extends TextView instead of View
- setTextIsSelectable(true), unfortunately this is >= API 11
- SetText("This is a test !")

In ConsoleActivity
- Remove all long click references (gestures & context menu).

And it seems work perfectly !
Obviously this still need work, but what do you think ?


output

PR #127, opened 2015-08-11
Use unicode symbols for arrow buttons

kruton:
This relieves some need for translations to many languages.


kruton:
@alescdb What do you think of this?

alescdb:
I haven't thought of unicode symbols !
It looks very nice on my N5 !

I've tried drawing arrows in Photoshop, but I'm definitely not an artist (even for an arrow symbol) :-)

Just wondering, on custom ROM when users use a theme font (or I think Samsung allow the user to change his font), all fonts may not have all unicode symbols. We may have to force the font ?

kruton:
The font selector should choose DroidSansFallback which should be present in all devices or else a lot of things will be broken.

PR #129, opened 2015-08-13
Native Copy/Paste (POC, not for merge)

alescdb:
Since I've accidentally closed Pull #126 where I've started this "experiment" I'm reopening a new pull with this patch (If there is another place to discuss this, plz tell me).

This patch (again, sorry, Android Studio messed up all Tab/Space) extends TerminalView with TextView. The TextView itself isn't displayed since TerminalView override onDraw(), so we just need to sync VDUBuffer with the TextView to allow native copy/paste while user perform a long clic (performLongClick).

Here is a short video of what I've done (for debuging purpose I've called super.onDraw() to see the TextView in green) :

output

CC: @jklein24 @bengolds


jklein24:
This looks really awesome, @alescdb! It seems that the diff didn't quite work out cleanly so it's a bit hard to follow what functionality you've removed here. What exactly are the tradeoffs? Anyway, nice work :-)

jklein24:
FYI - travis is angry about the spaces for indents instead of tabs. I actually prefer spaces, but for consistency reasons, make sure you switch back to tabs. This will also probably make the diff cleaner.

alescdb:
I've discibe what I've done in my last post in pull #126. It's very basic and simple, 4 lines of code :-)
In this pull I've just add the sync of VDUBuffer with base class TextView and made them superpose correctly.

I'll try to make a cleaner patch tomorrow so you can see more clearly.

jklein24:
Cool, so basically the main thing we lose is the ContextMenu with paste? Doesn't seem like the end of the world to me. Seems like a worthwhile tradeoff.

leonerd:
My $0.02 (AUD currently since I'm visiting Australia ;) ) is that I quite like it. I might even consider such a mode for pangoterm sometime as well.

app/src/main/java/org/connectbot/ConsoleActivity.java
```diff
@@ -563,17 +564,21 @@ public boolean onTouch(View v, MotionEvent event) {
 						MotionEventCompat.getSource(event) == InputDevice.SOURCE_MOUSE &&
 						event.getAction() == MotionEvent.ACTION_DOWN) {
 					switch (event.getButtonState()) {
-					case MotionEvent.BUTTON_PRIMARY:
-						// Automatically start copy mode if using a mouse.
-						startCopyMode();
-						break;
+						case MotionEvent.BUTTON_PRIMARY:
+							// Automatically start copy mode if using a mouse.
+							startCopyMode();
```
jklein24: Not sure this still makes sense... Do you have a bluetooth mouse you can test with? If not, I can take a look into mouse functionality after this lands.

app/src/main/java/org/connectbot/ConsoleActivity.java
```diff
@@ -482,14 +481,15 @@ public void onMenuVisibilityChanged(boolean isVisible) {
 			}
 		});
 
+
 		// detect fling gestures to switch between terminals
 		final GestureDetector detect = new GestureDetector(this, new GestureDetector.SimpleOnGestureListener() {
 			private float totalY = 0;
 
 			@Override
 			public void onLongPress(MotionEvent e) {
```
jklein24: Let's go ahead and kill this function.

app/src/main/java/org/connectbot/ConsoleActivity.java
```diff
@@ -308,7 +307,7 @@ public void onCreate(Bundle icicle) {
 		inflater = LayoutInflater.from(this);
 
 		pager = (ViewPager) findViewById(R.id.console_flip);
-		registerForContextMenu(pager);
+		// registerForContextMenu(pager);
```
jklein24: Feel free to totally remove the context menu in this CL if it doesn't make sense with this UX. I think we can figure something out for paste that works nicer with this flow.

app/src/main/java/org/connectbot/ConsoleActivity.java
```diff
@@ -585,55 +590,49 @@ public boolean onTouch(View v, MotionEvent event) {
 					SelectionArea area = copySource.getSelectionArea();
```
jklein24: Any reason not to totally get rid of all the selectionArea/copysource stuff now?

jklein24:
FWIW, I'm totally in favor of getting this landed once it's ready. I think it's a big UX improvement. @kruton, what are your thoughts?

jklein24:
@alescdb do you might syncing up and rebasing to resolve merge conflicts?

jklein24:
I patched this in to play with it and although I think this is a good approach, I ran into some problems:

- Font size needs to match the current user font size, otherwise everything overlaps and isn't readable. 
- The left selection anchor isn't visible on the left edge of the screen. Not sure what the right solution is here... You'd think TextView would account for this.
- The green text never goes away after selecting, so you can't really see what you're doing after that.
- The "setTextIsSelectable" function is only supported on API version 11+. Right now we support 8+.
- Scrolling if you're in the select mode makes everything go crazy.

We need to figure out a good strategy for *leaving* selection mode. Maybe as soon as the user starts typing?

bengolds:
For leaving select mode, I'd say: hitting the back button in the top left, hitting the system back button, hitting escape, hitting copy, or starting to type.

alescdb:
@jklein24 the green text is only here for debugging purpose to see the sync between VDUBuffer and the TextView and if text superpose well (see performLongClick) it's not supposed to be displayed at all. To get rid of it, just comment super.onDraw() in TerminalView.onDraw().
To avoid API 11+ dependency, it may be possible to replace TextView by an EditText (I haven't tried).
A better approach (but maybe less simpler) would be to create an TerminalView Editor based on android.widget.Editor.

jklein24:
re: "the green text is only here for debugging purpose to see the sync between VDUBuffer and the TextView and if text superpose well"

Yes, I understand this and agree that the green text should only be there for debugging, but my point is more that the text *doesn't* align in lots of cases. I see that maybe the scrolling isn't a problem because the alignment will get reset on the next long press, but the other issues still seem valid (left anchor, font size, etc.)

alescdb:
I recently switched my Nexus 5 to Android M preview 3, and this does not seems to work on M !
I don't understand why for now. This may only be a M bug.
Maybe we should wait for M release before spending more time into this approach.

jklein24:
Weird, I had it working on my Nexus 9 on M, but don't have M on my Nexus 5. What do you see going wrong?

alescdb:
Long press is detected and TextView is correctly overlapping, but Copy anchors are not displayed and text is not selected.

To answer your previous questions, for font size it should not be too hard to sync font size and line spacing.
For left anchors (and certainly the right anchor when at the end of the line) I see no other solution than adding a margin to TerminalView.

jklein24:
Closing in favor of #277

PR #133, opened 2015-08-15
Hide ActionBar when a virtual key is pressed.

alescdb:
When actionbar is setup to hide it's overlapping the console and we can't see what we are doing when pressing a virtual key.


alescdb:
This pull req is no longer necessary since it is included in pull req #134

Issue #135, opened 2015-08-16
Closing a connection does not go back to main screen (on git-master-2015-08-14_15-47-46)


alescdb:
When I close a single connection with the shell exit command on tablet (Nexus 9) the terminal screen stay open and a unique navigation tab is showed with the name of my ex-connection. The "Home" button in action bar is not displayed, so it impossible to go back to the connection screen.
When two connections are openned, closing a connection with exit cmd, the tab does not close.

This seems to be related to @jklein24 Tab addition.

Also, the tabs are centered and look weird and the "undescore" (selected tab) does not display until I change tab or swipe.

jklein24:
> When I close a single connection with the shell exit command on tablet (Nexus 9) the terminal screen
> stay open and a unique navigation tab is showed with the name of my ex-connection. The "Home" 
> button in action bar is not displayed, so it impossible to go back to the connection screen.
> When two connections are openned, closing a connection with exit cmd, the tab does not close.

I'm trying to repro this, but no luck so far. Here's what I'm doing:
1) Open up two connections.
2) Use "exit" close one. That one is closed and only the other is shown with no tabs.
3) Use "exit" to close the other remaining terminal. Closes and goes back to the hostlist as expected.

> Also, the tabs are centered and look weird and the "undescore" (selected tab) does not display until I change tab or swipe.

I've seen this once or twice, but I'm having trouble getting it to happen consistently. Do you have a good repro? Otherwise I'll just keep digging to see if I can track something down.

alescdb:
I think this has something to do with the "Close on disconnect" in the connection settings, I have FCs when I check this.
For the centered tab, I always have this, landscape and portrait as soon as I have more than one connection.

jklein24:
I don't think the tab centering is really a bug. That's just the default positioning for tabs. I can move them over, though. "Close on disconnect" seems like a likely candidate. If I enable that for both of my hosts, I can repro the crash. Without it, everything seems normal.

alescdb:
I'm not sure yet, but It seems the activity is completely locked (but no FC) only with an ssh connection with "Close on disconnect" checked.
Tabs should not be centered, usually they fill parent width.

jklein24:
> I'm not sure yet, but It seems the activity is completely locked (but no FC) only with an ssh connection with "Close on disconnect" checked.

I think I tracked this down. It looks like there's a codepath a missed when trying to make sure all the disconnect listeners were triggered on the UI thread. I'll send out a PR shortly.

[...]

PR #136, opened 2015-08-17
Handle immediate disconnection listeners on the UI thread

jklein24:
This will resolve a crash when disconnecting hosts with "close on disconnect" set to true. See https://github.com/connectbot/connectbot/issues/135.


jklein24:
@alescdb, this resolves the FC for me with "close on disconnect". Are you still seeing any issues with this?

alescdb:
No more FC and no more lock on empty activity :-)

jklein24:
Thanks!

PR #141, opened 2015-08-20
travis: make sure we do not increase lint issues

kruton:


kruton:
@jklein24 so we don't increase the number of issues :)

jklein24:
Cool!

PR #152, opened 2015-08-25
Travis: try to cache Android NDK to speed up build

kruton:


kruton:
This is too big for Travis's cache.

Issue #153, opened 2015-08-25
Rotating sometimes switches the active tab


jklein24:
It looks like the requested index sticks around when the device is rotated so tab selection always wants to revert back to the original selection for which the ConsoleActivity was launched.

PR #155, opened 2015-08-26
Save the currently visible connection between screen rotations.

jklein24:


jklein24:
Fixes https://github.com/connectbot/connectbot/issues/153

PR #156, opened 2015-08-27
Remove reference to grantpt

kruton:
grantpt doesn't do anything on Android since it uses devpts.


kruton:
@jklein24 this should fix the problem with older devices.

PR #160, opened 2015-08-29
Fix race condition in updating the HostListActivity with current host statuses

kruton:


kruton:
@jklein24 this fixes the test failure in the new test for me.

Issue #166, opened 2015-08-30
Please make buttons on virtual keyboard light up when being pressed (for a visual feedback)


wp9015362:
Hello,

could you please make the buttons on the virtual keyboard light up when they are being pressed?

That way there would be a visual feedback and the user would be able to see that the button actually has been pressed.

Currently there is no visual feedback at all when the buttons are being pressed, which results in a rather poor UX IMHO.

Could you please do something about it?

Regards

wp9015362:
PS:

IMHO there should also be some kind of visual indication that if the CTRL button is being pressed on the virtual keyboard, that it kinda remains pressed until you press another button like the "C" key for example.

Because, as far as I can tell, if you want to press CTRL + C for example, then you need to tap the CTRL button, but you don't need to hold it because it remains being pressed until you tap on the C key. Maybe I'm wrong, but this is how it seems to work.

And therefore there also should be some kind of visual indication so that the user knows that if he presses the CTRL button, that it will remain pressed until another button has been pressed.

Regards

kruton:
The buttons light up when they are pressed now, but "Ctrl" does not show it is pressed yet. There is visual indication in the cursor indicator (this has been there since 2009 or so).

nindanaoto:
This seems to be fixed in the main.

PR #178, opened 2015-08-31
Provide visual feedback for virtual keyboard

alescdb:
Feature req : #166 

PS: I have trouble with virtual keyboard on M, the bell repeat a random number of time. I've checked logcat and all keyboard calls in ConsoleActivity are called only once. I suspect it happens in vt320.keyPressed(), but I haven't traced it yet.


kruton:
@alescdb do you want to try to write a instrumentation test for this? I just added StartupTest.java at [`app/src/androidTest/java/org/connectbot/StartupTest.java`](https://github.com/connectbot/connectbot/blob/master/app/src/androidTest/java/org/connectbot/StartupTest.java)

PR #180, opened 2015-09-03
Try to show new lint issues in check-lint-count.sh

alescdb:
I'm always lost with new lint issues... so the idea here is to store the last good lint results and to compare it with the new one, and show the diff if count is higher.

### DIFF
--- a/scripts/check-lint-count.sh
+++ b/scripts/check-lint-count.sh
@@ -4,10 +4,6 @@
 # Travis CI builds to fail when the number increases by exploiting the
 # caching mechanism.
 
-# This is to prime the system: when I submitted this change, this is the
-# number of lint warnings that existed.
-DEFAULT_NUMBER=207
-
 if [[ $# != 2 || ! -f $1 ]]; then \
     echo "Usage: $0 <lint.xml file> <historical.xml file>"
     exit 1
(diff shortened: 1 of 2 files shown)


scripts/check-lint-count.sh
```diff
@@ -8,11 +8,8 @@
 # number of lint warnings that existed.
 DEFAULT_NUMBER=207
```
kruton: Should delete this if you're not using it.

scripts/check-lint-count.sh
```diff
@@ -27,21 +24,31 @@ if [[ ! -x $xmllint ]]; then \
...
-else \
-    historical_count=$DEFAULT_NUMBER
+if [[ ! -f $2 ]]; then \
+  # no cache history, store this one and exit
+  cp $1 $2
+  exit 0
 fi
 
-new_count="$($xmllint --xpath 'count(//issue)' "$lint_file")"
+echo "cat //issue/location" | xmllint --shell $historical_file | grep '<location' >/tmp/hist.$$
```
kruton: How about:
```
tmp_dir="$(mktemp -d lint.XXXXXXXX)"
trap "rm -rf $tmp_dir" EXIT ERROR
lint_file="$tmp_dir/lint.xml"
hist_file="$tmp_dir/hist.xml"
```
Then you don't need to `rm -f` the files everywhere else.

kruton:
We mostly care about `EXIT`. I suppose `ERR` is only for when there is `set -e` and signals are only for when the script itself is sent a signal via `kill(...)`.

PR #182, opened 2015-09-03
Fix JaCoCo usage with connected tests

kruton:


kruton:
Poo, this apparently breaks Coveralls.

PR #185, opened 2015-09-04
Style the host list items with material design.

### LINKED ISSUE #175: Font size in list of hosts is way too big
Hello,

the font size in the list of hosts is way too big.

This becomes especially noticeable when inreasing the Android system font size above the standard value, because then it becomes really huge and the hostnames even become cut off.

Other apps that I know of are using a more reasonable font size.

So, could you please reduce the font size in the list of hosts?

Hint: The font size that is being used for the settings list is much better.

You might also want to have a look at the guidelines for two-line lists in the Google Material Design guidelines over there:

https://www.google.com/design/spec/components/lists.html

Regards

khorimoto:
This change updates the items to be in line with the guideline here: https://www.google.com/design/spec/components/lists.html#lists-specs.

Closes #175.


khorimoto:
screenshot from 2015-09-04 13 43 58

PR #187, opened 2015-09-04
Add .idea/gradle.xml to .gitignore files and add app/app.iml changes.

khorimoto:
These files change according to the individual user's machine each time the project is edited by Android Studio.


kruton:
`gradle.xml` is fine, but I don't think `app/app.iml` should be ignored.

PR #189, opened 2015-09-05
Allow temporary cloned connections

alescdb:
I often use more than one connection to a server (sometimes even more than 2). So, in the actual implementation of ConnectBot, we have to create a second host with the same specifications (without any way to clone an existing host).

This patch add a context menu in the hosts list "Clone connection" that will clone the HostBean in memory and change the nickname to "My Server #1",  and connect to it.

~~There~~ ~~is~~ ~~a~~ ~~flaw~~ ~~here,~~ ~~the~~ ~~temporary~~ ~~status~~ ~~is~~ ~~based~~ ~~on~~ ~~the~~ ~~host~~ ~~nickname~~ ~~(regex~~ ~~".*~~ ~~#[0-9]+$"),~~ ~~so~~ ~~if~~ ~~a~~ ~~user~~ ~~has~~ ~~defined~~ ~~manually~~ ~~a~~ ~~nickname~~ ~~ending~~ ~~with~~ ~~this~~ ~~pattern~~ ~~it~~ ~~will~~ ~~be~~ ~~detected~~ ~~as~~ ~~temporary.~~ ~~The~~ ~~only~~ ~~consequence~~ ~~will~~ ~~be~~ ~~that~~ ~~the~~ ~~user~~ ~~won't~~ ~~see~~ ~~the~~ ~~"Clone~~ ~~connection"~~ ~~context~~ ~~menu.

~~I'm~~ ~~not~~ ~~sure~~ ~~this~~ ~~is~~ ~~the~~ ~~proper~~ ~~way~~ ~~to~~ ~~implement~~ ~~this.~~

### DIFF
--- a/app/src/main/java/org/connectbot/ConsoleActivity.java
+++ b/app/src/main/java/org/connectbot/ConsoleActivity.java
@@ -182,7 +182,7 @@ public class ConsoleActivity extends AppCompatActivity implements BridgeDisconne
 				}
 			}
 
-			if(requestedBridge != null) {
+			if (requestedBridge != null) {
 				Intent intent = getIntent();
 				if (intent != null) {
 					requestedBridge.setTemp(intent.getBooleanExtra(IS_TEMPORARY, false));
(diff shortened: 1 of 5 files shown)


jklein24:
I really like this feature, but I'll leave the final say to @kruton :-)

JohnVeness:
I haven't tested this, but I too would find this useful on occasion.

PR #205, opened 2015-09-15
Changed Settings options to use switches instead of checkboxes.

### LINKED ISSUE #169: ConnectBot should use switches instead of checkboxes to adhere to Google Material Design Guidelines
Hello,

currently connectbot is using checkboxes in the settings.

But the Google Material Design Guidelines clearly say:

> https://www.google.com/design/spec/components/selection-controls.html
> 
> [...] If you have a single option, avoid using a checkbox and use an on/off switch instead. [...]

So, could you please replace them with switches?

Regards

rhansby:
Fixes #169.


kruton:
Can you attach a screenshot?

rhansby:
screenshot-switches-settings

alescdb:
Just a warning, SwitchPreference is API level 14 and higher
http://developer.android.com/reference/android/preference/SwitchPreference.html

rhansby:
screenshot-settings

kruton:
Yes, we are still API 8 and up. Check out [this post on StackOverflow](http://stackoverflow.com/a/31229920) for a possible solution.

rhansby:
When the app is run on devices pre-APIv14, should the Preferences be Checkboxes, and then v14 and above should show Switches?

Or should Switches be shown on all devices, regardless of the API version they have?

kruton:
It would be nice to have a consistent app as much as possible. It appears the support library has a Switch in it but not the SwitchPreference.

rhansby:
Added switches to pre-APIv14 using SwitchCompat:

screenshot-10-settings

rhansby:
Using SwitchCompact on preAPIv10 seems to have a visual bug, shown below. Does anyone know what might be causing this?

output

wp9015362:
> @rhansby wrote:
> 
> Using SwitchCompact on pre-APIv14 seems to have a visual bug, shown below (the bottom switch animates when not touched). Does anyone know what might be causing this?

Today I came across the following article:

http://www.androidauthority.com/android-design-support-library-overview-636694/

Is it any helpful?

Also:

The pubkey generation view is also affected. The pubkey generation view has two checkboxes. They should also be changed to switches to adhere to the Google Material Design Guidelines, see following issue for a screenshot:

https://github.com/connectbot/connectbot/issues/208

Regards

rhansby:
Thanks for the article @wp9015362. After quickly glancing over it I unfortunately don't think that will help, but I'm continuing to look into it :)

rhansby:
Regarding your pubkey generation view comment, I think that checkboxes are still appropriate, given their slightly different context. However I want @kruton to give the final say on that. In any case thank you for pointing that out.

wp9015362:
> @rhansby wrote:
> 
> Regarding your pubkey generation view comment, I think that checkboxes are still appropriate, given their slightly different context. However I want @kruton to give the final say on that.

Why is it a different context?

Well, let's see what the Google Material Design Guidelines have to say about it:

> https://www.google.com/design/spec/components/selection-controls.html
> 
> [...] If you have a single option, avoid using a checkbox and use an on/off switch instead. [...]

:wink:

So, since those two checkboxes on the pubkey generation view also each are for single options, they should be changed to switches as far as I can tell :yum:.

It's the same as with the settings view.

By the way: I simply haven't noticed the two checkboxes on the pubkey generation view earlier because of the following issue:

https://github.com/connectbot/connectbot/issues/212

Otherwise I would have included them in https://github.com/connectbot/connectbot/issues/169 from the beginning.

Regards

rhansby:
The most recent push is a workaround for the visual glitch noted above. See GIF for current (bug-free) behavior on pre API-v14:

output

wp9015362:
@rhansby :

The edit host view is also affected. You can reach the edit host view by tapping and holding on a host in the host list and then selecting "Edit host".

The edit host view has four checkboxes which also should be changed to switches to adhere to the Google Material Design Guidelines.

Regards

rhansby:
@wp9015362 , this PR also affected the host preferences list :)

PR #206, opened 2015-09-15
Host list: convert to RecyclerView.

khorimoto:
This converts the rest of the view to the Material Design spec. Changes include item animation as well as the "shadow" semicircles which appear at the top and bottom of the list when scrolling. Furthermore, performance should be improved through automatic view recycling.

Note that the click handler and context menu handler were only moved and were left unchanged. RecyclerView use requires that these handlers are applied to the individual child views instead of being added to the list ViewGroup as a whole.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -59,6 +59,7 @@ android {
     }
 
     dependencies {
+        compile 'com.android.support:recyclerview-v7:+'
         compile "com.android.support:support-v4:23.0.1"
         compile "com.android.support:appcompat-v7:23.0.1"
         compile "com.android.support:design:23.0.1"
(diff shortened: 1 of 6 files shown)


khorimoto:
screenshot from 2015-09-16 11 49 59

kruton:
@khorimoto can you get `./gradlew check` happy?

khorimoto:
Done.

kruton:
Whoops, looks like this caused some errors:

```
org.connectbot.StartupTest > localConnectionCanDelete[Nexus 7 - 5.1]  [31mFAILED  [0m
        java.lang.IllegalStateException: You need to use a Theme.AppCompat theme (or descendant) with this activity.
        at android.support.v7.app.AppCompatDelegateImplV7.createSubDecor(AppCompatDelegateImplV7.java:309)
```

Not sure if this is a real error, but might be if the ID was changed:
```
org.connectbot.StartupTest > localConnectionCanChangeToRed[SK17a - 2.3.3]  [31mFAILED  [0m
        android.support.test.espresso.NoMatchingViewException: No views in hierarchy found matching: with id: android:id/list
        If the target view is not part of the view hierarchy, you may need to use Espresso.onData to load it from one of the following AdapterViews:android.support.v7.widget.AppCompatSpinner@2affa350
```

PR #211, opened 2015-09-16
Use 4 spaces per tab instead of 8.

khorimoto:
This makes lines shorter and easier to read.


kruton:
This doesn't appear to change the tab width in the UI, it just makes it use 4 spaces instead of tab characters.

khorimoto:
Okay, changed it to use tabs by default but keep the tabs to be equal to 4 spaces. The tab width in the UI is noticeably improved.

PR #218, opened 2015-09-17
Fix StartupTest tests to work with RecyclerView.

khorimoto:
The tests were broken after I converted from ListView to ReyclerView. There is still one issue left: localConnectionDisconnectConsoleActivity() still fails, and, ironically, it is the only test that did not need to be modified. @kruton, would you mind taking a look at this for me? Thanks.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -76,6 +76,12 @@ android {
         androidTestCompile('com.android.support.test.espresso:espresso-intents:2.2') {
             exclude module: "support-annotations"
         }
+        androidTestCompile ('com.android.support.test.espresso:espresso-contrib:2.2'){
+            exclude module: 'support-annotations'
+            exclude module: 'support-v4'
+            exclude module: 'support-v13'
+            exclude module: 'recyclerview-v7'
+        }
     }
 
     buildTypes {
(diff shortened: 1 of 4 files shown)


kruton:
Yes, I'll take a look at it. The one thing I questioned was whether these tests were meant to focus on the View themselves instead of the underlying data structure. It seems like it should focus more on the View level instead of the ViewHolder's contents.

kruton:
It might be happening if you have your animation scale in developer options set to anything other than "none."

khorimoto:
Regarding whether this should be focused on the View vs. ViewHolder: I tried to copy the original tests as closely as possible, and these tests used onData() to fetch based on the HostBean instead of using onView(). Are you saying that you would prefer this no longer to be the case?

Regarding the animation scale, how do we set this for tests? And why is this appearing now when it was not there before with the same test code?

kruton:
If you focus on the View then you don't need the withHolder method. It just
seems like the Espresso API is trying to get you to focus on integration
tests where you try to focus on the highest level abstraction. This might
already be the case in your change.

Also Espresso seems to be racy sometimes. The test that was failing for you
failed half the time with animations turned on. They are enabled in
Settings -> Developer settings.

khorimoto:
Regarding focusing the View: if we don't use the ViewHolder method, we would need to choose the 0th child view of the RecyclerView and just assume that it is the correct one. Is this assumption okay to have? Seems to me that it breaks abstraction.

kruton:
Yes, I think it's fine the way it is. I was initially thinking you were going to look at `holder.host` which is not the case here.

PR #219, opened 2015-09-18
Use FloatingActionButton to add new hosts to the host list.

khorimoto:
This updates the view to be material design themed.


app/src/main/res/values/colors.xml
```diff
@@ -27,4 +27,9 @@
 	<color name="primary">#00BCD4</color>
 
 	<color name="accent">#F4511E</color>
+        <color name="dark_accent">#BF360C</color>
```
kruton: These ended up with spaces instead of tabs.

kruton:
Is this going to continue on to having a new screen to add hosts? I'd like to get rid of the current `HostEditorActivity` and have something with a better layout that can be used for both adding and editing existing hosts.

app/src/main/res/values/colors.xml
khorimoto: Fixed.

khorimoto:
The dialog I added in this change is the new screen to add hosts. Did you want something different?

kruton:
It's just the old spinner + hostname field in a dialog which I never particularly thought was very good looking. If we're popping something up to the user, we might as well allow them to edit all the fields they want up front.

Also, this will break the instrumentation tests in `app/src/androidTest`.

PR #229, opened 2015-09-22
added pipe, brackets, braces and back slash to the popup button list.

nixomose:
I love connectbot and I do lots of system administration with it all the time, but my hardware keyboard lacks any way to send a pipe key, so I have to close the keyboard, open the popup keyboard hit a few buttons then open the keyboard again just to get the pipe symbols.

This pull just adds | [ ] { } and \ to the little popup button list making these keys easier to enter.


jklein24:
@nixomose, do you mind rebasing and resolving any merge conflicts?

nixomose:
all rebased and merged and all that.

jklein24:
@bengolds @kruton, what are your thoughts on this? The virtual keys bar is getting a bit crowded, but I do see the value here. I'm somewhat torn.

rhansby:
Perhaps we could make this feature an optional setting?

bengolds:
Since these keys are already provided on the normal virtual keyboard and this just makes them a little easier to get to, I'd rather we make this an advanced feature.

PR #240, opened 2015-09-23
Fixed error when not filling in all fields of port forward creation dialog

### LINKED ISSUE #236: Unreported/unexplained error/failure during portforward creation
If one only enters the desired name of a new port forward (screenshot below) and presses "Create Port Forward", the port forward is not created, and the app does not inform the user that creation failed or why.

Partial stack trace:

    09-22 17:01:27.416  27549-27549/org.connectbot.debug E/CB.PortForwardListAct﹕ Could not update port forward
    java.lang.NumberFormatException: Invalid int: ""
            at java.lang.Integer.invalidInt(Integer.java:138)
            at java.lang.Integer.parseInt(Integer.java:358)
            at java.lang.Integer.parseInt(Integer.java:334)
            at org.connectbot.bean.PortForwardBean.<init>(PortForwardBean.java:72)
            at org.connectbot.PortForwardListActivity$2$2.onClick(PortForwardListActivity.java:191)
            at com.android.internal.app.AlertController$ButtonHandler.handleMessage(AlertController.java:163)

portforward bug

rhansby:
fixes #236

### DIFF
--- a/app/src/main/java/org/connectbot/PortForwardListActivity.java
+++ b/app/src/main/java/org/connectbot/PortForwardListActivity.java
@@ -188,11 +188,23 @@ public class PortForwardListActivity extends AppCompatListActivity {
 										break;
 									}
 
+									// Why length(), not isEmpty(), is used: http://stackoverflow.com/q/10606725
+									String sourcePort = sourcePortEdit.getText().toString();
+									if (sourcePort.length() == 0) {
+										sourcePort = sourcePortEdit.getHint().toString();
+									}
+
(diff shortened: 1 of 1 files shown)


jklein24:
This seems reasonable to me (default to the hint value). Couldn't there be other invalid values besides empty though? What happens then? We may want to do some more validation.

rhansby:
I also considered that, but the "source port" field only accepts integers, and an incorrect destination field is the users fault and cannot be caught.

PR #243, opened 2015-09-23
Make HostListActivity use intended title 'Hosts'.

### LINKED ISSUE #242: Toolbar still says "ConnectBot" in host list view. Should say "Hosts" instead.
@rhansby :

In https://github.com/connectbot/connectbot/issues/231 you said:

> We could remove the blurb altogether, but then it isn't at all clear that we are on the HostList view.

That's mainly because the toolbar only says "ConnectBot" in the host list view.

If you would replace "ConnectBot" with "Hosts", that would instantly solve the issue mentioned above.

Actually I had requested that in https://github.com/connectbot/connectbot/issues/171 and @kruton said it would have been fixed with https://github.com/connectbot/connectbot/commit/3ee07166ee1aa584686b631341cb5dabc32359e5, but it hasn't, the toolbar still says "ConnectBot" instead of "Hosts" (according to your screenshot in https://github.com/connectbot/connectbot/issues/231).

Yes, I realize that the host list view basically is the main screen / home screen of ConnectBot and that the toolbar saying "ConnectBot" might actually be intentional because of that.

But if you look at other Material Design apps, a lot of them are showing the actual view name on the toolbar instead of the app name, even on their main screen / home screen.

So, I'd say that would also make sense for ConnectBot.

On all other views in ConnectBot, the toolbar names the proper view, i.e. in the settings the toolbar says "Settings", on the help view it says "Help" and so on...

So on the host list view the toolbar should also do the same and say "Hosts".

Any chance you could do that?

Regards

rhansby:
Fixes #242

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -44,6 +44,7 @@
 
 		<activity
 			android:name=".HostListActivity"
+			android:label="@string/title_hosts_list"
 			android:launchMode="singleTop">
 			<intent-filter>
 				<action android:name="android.intent.action.MAIN"/>


kruton:
This makes the application appear as "Hosts" in the list of all apps.

PR #246, opened 2015-09-23
Add padding to bottom of HostList to prevent FAB covering list.

### LINKED ISSUE #241: Floating action button is hidding connection status
screenshot_20150923-204501

rhansby:
Fixes #241 

Before: 
fab before

After:
fab after

### DIFF
--- a/app/src/main/res/layout/act_hostlist.xml
+++ b/app/src/main/res/layout/act_hostlist.xml
@@ -26,10 +26,14 @@
 	android:orientation="vertical"
 	>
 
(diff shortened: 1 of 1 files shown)


rhansby:
@khorimoto, please be aware of this as you address @168 and adding FABs in general.

jklein24:
My only concern is whether this would look bad on a really small screen. Can you test on an old emulator with a smaller height?

rhansby:
It seems to look fine on older devices/OSs. See the below screenshot running on APIv10.

It doesn't perfectly match row height as in newer devices, but still looks good.

old device

(this screenshot also reveals an issue I have documented at #247)

PR #251, opened 2015-09-24
Only use host if it is not null

kruton:
Since we null check the host, we should make sure not to dereference it
later on.


jklein24:
Out of curiosity, under what circumstances is the host null?

kruton:
I'm not sure that it could be, but might be changed on another thread.

PR #252, opened 2015-09-24
Create EditHostActivity.

khorimoto:
EditHostActivity is the new host creation/editing activity which will eventually replace HostEditorActivity.

This PR includes:
1) Skeleton for the activity.
2) Skeleton for the whole editor fragment.
3) Complete implementation for the URI editor fragment, which is included from the editor fragment.

The URI editor fragment allows the user to input a "quick connect" URI or to edit the URI piece by piece.

None of this code is integrated/accessible from in the app yet. This will occur in a future PR.


jklein24:
Nice! cc @kruton for help on this review, but i'll take a look too.

app/src/main/AndroidManifest.xml
```diff
@@ -85,6 +85,16 @@
 		</activity>
 
 		<activity
+				android:name=".EditHostActivity"
+				android:configChanges="keyboardHidden|orientation"
```
jklein24: Does this activity need to handle this itself?

khorimoto: Removed.

khorimoto:
I checked out the lint errors, and the majority of them were not caused by this PR. The only ones I could determine were caused by this CL is that I included the animateLayoutChanges tag in my XML, which isn't available before API v11. However, the documentation says, "this is not an error; the application will simply ignore the attribute," so I don't think there's any problem here. Let me know if I'm wrong.

khorimoto:
Preview: https://goo.gl/photos/tFjeBKjVHzfn8rhX9

wp9015362:
> @khorimoto wrote:
> 
> Preview: https://goo.gl/photos/tFjeBKjVHzfn8rhX9

No offence, but, I don't like that there is a ***username@hostname:port*** field and that you have to click the expand button first to see the other separate fields.

IMHO the ***username@hostname:port*** field is redundant and unintuitive (why would I want to type an @-sign and a colon, if I could just jump to the next field instead, which is easier on a touchscreen IMHO).

I think the ***username@hostname:port*** field should be removed and the separate ***Username***, ***Host*** and ***Port*** fields should directly be expanded instead.

Also... minor issue, but :yum:: In the ***username@hostname:port*** field it's "hostname", whereas in the separate field it's just "Host", a little inconsistent. But that issue would also be solved by removing the ***username@hostname:port*** field :yum:.

Regards

khorimoto:
> No offence, but, I don't like that there is a username@hostname:port field and that you have to click the expand button first to see the other separate fields.

Thanks for the suggestion, but I disagree. We want to make it easy for people to SSH the way they normally would (by typing the URI like someone would on a terminal). This also goes in line with other apps; for example, when adding a contact with the Contacts app that ships with Android, you can enter a full name or click an expand button to reveal fields for individual parts of the name (first name, middle name, last name, etc).

wp9015362:
@kruton :

Well, you might have a point there.

Another thing:

May I ask how you are planning to integrate the other remaining host editor fragments?

Will you put them all on one view, just like it is on the current Edit Host view?

Or will you separate the URI fragment from the other fragments with tabs:

https://www.google.com/design/spec/patterns/navigation.html

?

Regards

app/src/main/res/layout/fragment_host_editor.xml
```diff
@@ -0,0 +1,137 @@
...
+        android:background="#00000000" />
+
+  </LinearLayout>
+
+  <LinearLayout
+      android:id="@+id/uri_parts_container"
+      android:layout_width="match_parent"
+      android:layout_height="wrap_content"
+      android:orientation="vertical"
+      android:layout_marginLeft="56dp"
```
kruton: Forgot a `layout_marginStart` here.

app/src/main/res/layout/activity_edit_host.xml
```diff
@@ -0,0 +1,23 @@
...
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
+	xmlns:tools="http://schemas.android.com/tools"
+	android:id="@+id/fragment_container"
+	android:layout_width="fill_parent"
+	android:layout_height="fill_parent"
+	tools:context="org.connectbot.EditHostActivity" />
```
kruton: add: `tools:ignore="MergeRootFrame"`

app/src/main/res/layout/fragment_host_editor.xml
```diff
@@ -0,0 +1,137 @@
...
+        android:id="@+id/transport_selector"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content" />
+
+  </LinearLayout>
+
+  <LinearLayout
+      android:layout_width="match_parent"
+      android:layout_height="wrap_content"
+      android:animateLayoutChanges="true">
```
kruton: add `tools:ignore="UnusedAttribute"`

app/src/main/res/layout/fragment_host_editor.xml
```diff
@@ -0,0 +1,137 @@
...
+  </LinearLayout>
+
+  <LinearLayout
+      android:id="@+id/uri_parts_container"
+      android:layout_width="match_parent"
+      android:layout_height="wrap_content"
+      android:orientation="vertical"
+      android:layout_marginLeft="56dp"
+      android:visibility="gone"
+      android:animateLayoutChanges="true">
```
kruton: add `tools:ignore="UnusedAttribute"`

app/src/main/res/values/strings.xml
```diff
@@ -594,4 +594,8 @@
 	<string name="button_key_f11">F11</string>
 	<!-- Text for the "F12" button in virtual keyboard. -->
 	<string name="button_key_f12">F12</string>
+	<string name="connection_type">Connection type:</string>
```
kruton: You need to add a comment above all these to give translators context for each string. They can't see the UI most of the time, so it has to describe what it's for and where it's used.

kruton: Oh, also this doesn't seem to be used anywhere, so remove it if so.

app/src/main/res/layout/fragment_host_editor.xml
khorimoto: Done.

app/src/main/res/layout/activity_edit_host.xml
khorimoto: Done.

app/src/main/res/layout/fragment_host_editor.xml
khorimoto: Done.

app/src/main/res/layout/fragment_host_editor.xml
khorimoto: Done.

app/src/main/res/values/strings.xml
khorimoto: Done.

khorimoto:
> Will you put them all on one view, just like it is on the current Edit Host view?

Aiming for all in one view.

PR #267, opened 2015-09-25
Forward mouseevents to the terminal when appropriate.

### LINKED ISSUE #225: Explore forwarding mouse events to connected host

jklein24:
Only handle mouse events locally if shift is held.
Partially fixes #225. Still need to figure out mouse wheel forwarding
and proper copy support when the terminal is handling selection itself
(ie in an editor).

### DIFF
--- a/app/src/main/java/de/mud/terminal/vt320.java
+++ b/app/src/main/java/de/mud/terminal/vt320.java
@@ -336,6 +336,10 @@ public void setScreenSize(int c, int r, boolean broadcast) {
     this(80, 24);
   }
 
+  public boolean isMouseReportEnabled() {
+    return mouserpt != 0;
+  }
+
   /**
    * Terminal is mouse-aware and requires (x,y) coordinates of
    * on the terminal (character coordinates) and the button clicked.
--- a/app/src/main/java/de/mud/terminal/vt320.java
+++ b/app/src/main/java/de/mud/terminal/vt320.java
@@ -385,26 +385,63 @@ public void setScreenSize(int c, int r, boolean broadcast) {
    * @param meta
    */
   public void mouseWheel(boolean down, int x, int y, boolean ctrl, boolean shift, boolean meta) {
-   if (mouserpt == 0 || mouserpt == 9)
-    return;
+    if (mouserpt == 0 || mouserpt == 9)
+      return;
 
-   int mods = 0;
-   if (ctrl) mods |= 2;
-   if (shift) mods |= 1;
-   if (meta) mods |= 4;
+    int mods = 0;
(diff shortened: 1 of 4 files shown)


jklein24:
cc @leonerd

jklein24:
Now this fixes #225

jklein24:
@kruton this should be good to go now.

app/src/main/java/org/connectbot/TerminalView.java
```diff
@@ -323,13 +326,29 @@ public boolean deleteSurroundingText (int leftLength, int rightLength) {
 	}
 
 	@Override
+	@TargetApi(12)
 	public boolean onGenericMotionEvent(MotionEvent event) {
+		if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.HONEYCOMB) return false;
```
kruton: I don't think it's possible to get here since this didn't exist in the parent class until API level 12 (honeycomb MR1)

app/src/main/java/de/mud/terminal/vt320.java
```diff
@@ -372,13 +376,81 @@ public void mousePressed(int x, int y, int modifiers) {
...
+  public void mouseWheel(boolean down, int x, int y, boolean ctrl, boolean shift, boolean meta) {
+    if (mouserpt == 0 || mouserpt == 9)
+      return;
+
+    int mods = 0;
+    if (ctrl) mods |= 16;
+    if (shift) mods |= 4;
+    if (meta) mods |= 8;
+
+    int mousecode = ((down ? 0 : 1) + 96) | 0x20 | mods;
```
kruton: Where does 96 come from? The spec says 64+{btn1,btn2} for wheel.

kruton:
LGTM besides the nit about seemingly unnecessary check.

app/src/main/java/org/connectbot/TerminalView.java
jklein24: Removed.

leonerd:
This seems good for a first cut, but sometime it might be nice to support the other reporting modes; or at the very least, the nice SGR-style one that xterm et.al. use; it has a multitude of nice properties and I fully expect more terminal-based programs to start making use of it.

jklein24:
@leonerd, I assume you mean these [extended coordinate modes](http://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h2-Extended-coordinates). Is 1006 what you're talking about? I suppose I could implement this, but it would take some more digging through vt320. How common is this use case? Do you have any programs I could use to test this mode?

PR #271, opened 2015-10-01
Add host display properties (e.g., color, font size) to add/edit host view.

khorimoto:
Note that the XML had 2-space tabs for some reason, so I fixed that as well. That's why it shows up as a bigger XML change than it actually was.

### DIFF
--- a/app/src/main/java/org/connectbot/HostEditorFragment.java
+++ b/app/src/main/java/org/connectbot/HostEditorFragment.java
@@ -19,6 +19,7 @@ package org.connectbot;
 
 import android.content.ContentValues;
 import android.content.Context;
+import android.content.res.TypedArray;
 import android.net.Uri;
 import android.os.Bundle;
 import android.os.Parcelable;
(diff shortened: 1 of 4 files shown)


jklein24:
Do you have a mock or spec that you're working off of for this view?

khorimoto:
Nope. @bengolds told me just to add a basic outline and he will give me a spec/mock at a later point. The UI is rough and unpolished for now.

app/src/main/java/org/connectbot/HostEditorFragment.java
```diff
@@ -234,6 +294,10 @@ public void onAttach(Context context) {
 		} catch (ClassCastException e) {
 			throw new ClassCastException(context.toString() + " must implement Listener");
 		}
+
+		// Now that the fragment is attached to an Activity, fetch the array from the attached
+		// Activity's resources.
+		mColorValues = getResources().obtainTypedArray(R.array.list_color_values);
```
jklein24: Should this TypedArray be recycled on detach?

khorimoto: Done.

PR #273, opened 2015-10-02
Add/edit host view: add fields for the connection-related preferences.

khorimoto:
This PR adds all the remaining UI fields and hooks them up to the fragment's HostBean object.

Remaining work includes:
  1) Passing in initial data to the fragment if it exists.
  2) Saving the host after it has been edited.
  3) UI polish (pending specs from @bengolds). The current design is very raw/ugly.

### DIFF
--- a/app/src/main/java/org/connectbot/HostEditorFragment.java
+++ b/app/src/main/java/org/connectbot/HostEditorFragment.java
@@ -156,6 +156,7 @@ public class HostEditorFragment extends Fragment {
 		for (int i = 0; i < transportNames.length; i++) {
 			if (transportNames.equals(mHost.getProtocol())) {
 				mTransportSpinner.setSelection(i);
+				break;
 			}
 		}
 		mTransportSpinner.setOnItemSelectedListener(new AdapterView.OnItemSelectedListener() {
--- a/app/src/main/java/org/connectbot/HostEditorFragment.java
+++ b/app/src/main/java/org/connectbot/HostEditorFragment.java
@@ -39,7 +39,6 @@ import android.widget.EditText;
 import android.widget.ImageButton;
 import android.widget.SeekBar;
 import android.widget.Spinner;
-import android.widget.Switch;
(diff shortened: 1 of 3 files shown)


jklein24:
Can you send a screenshot and fix those lint errors?

khorimoto:
Fixed lint errors. Screenshot below.

screenshot from 2015-10-05 16 47 40

app/src/main/java/org/connectbot/HostEditorFragment.java
```diff
@@ -128,12 +148,16 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
...
 		ArrayAdapter<String> transportSelection = new ArrayAdapter<>(
-				getActivity(),
-				android.R.layout.simple_spinner_item,
-				TransportFactory.getTransportNames());
+				getActivity(), android.R.layout.simple_spinner_item, transportNames);
 		transportSelection.setDropDownViewResource(android.R.layout.simple_spinner_dropdown_item);
 		mTransportSpinner.setAdapter(transportSelection);
+		for (int i = 0; i < transportNames.length; i++) {
+			if (transportNames.equals(mHost.getProtocol())) {
+				mTransportSpinner.setSelection(i);
```
jklein24: break;? Should be consistent one way or the other.

khorimoto: Done.

app/src/main/res/values/strings.xml
```diff
@@ -596,5 +596,6 @@
 	<string name="protocol_spinner_label">Protocol</string>
 	<!-- Label for button which expands/collapses section. -->
 	<string name="expand">Expand</string>
+	<string name="hostpref_authagent_with_confirmation">require confirmation</string>
```
jklein24: Can you add a comment above for context?

khorimoto:
Added string comment.

Issue #260, opened 2015-09-25
Terminal cursor does not adjust to extreme terminal size changes


wp9015362:
Hello,

when reducing the terminal size by a large amount and then increasing it again, the cursor is shifted to a few lines above (how much depends on how extreme the terminal size change was).

Here is how to reproduce it:

1. Activate "Volume keys change font size" in the ConnectBot settings
2. Connect to a host
3. Make sure the terminal is populated with a few lines of text (i.e. doing a *ls* or whatever), so that it is not empty
4. Now increase the font size via the volume up button on your device to as large as possible to reduce the terminal size to as small as possible (1 x 1 or something like that)
5. Now reduce the font size again via the volume down button on your device to make the terminal size increase to a moderate size again
6. As you can see, the cursor has now been shifted to a few line above and remains there

Could you please fix it?

Regards

wp9015362:
PS:

Pardon, I mixed up "reducing/increasing font size" and "reducing/increasing terminal size" a little bit. I have corrected it now.

Regards

jklein24:
I believe this is the same root cause as #80. We should really fix both. I'll leave this open until I can verify that it's a dupe.

PR #274, opened 2015-10-03
Fix bad math in moving the cursor when the window size changes.

### LINKED ISSUE #80: Cursor jumps to the middle of the screen after 2 rotations
In ConnectBot 1.8.4, after choosing "Automatic" in the "Rotation Mode" setting, you can start at the bottom of a portrait view like in step1.png, then rotate to landscape as in step2.png, and then rotate back to portrait.
You'll notice the cursor moves out of the original place in step 2, and "jumps" to the middle vertically in step 3.

step1
step2
step3

jklein24:
Fixes #80 

cc @rhansby


jklein24:
Also fixes #260

Issue #36, opened 2014-09-27
No ipv6 support


alive4ever:
Inserting a sequence of ipv6 address such as 2620:0:1cfe:face:b00c::3 will result in connectbot warning about malformed hostname. Putting the ipv6 address in brackets has no effect. Connectbot still treats ipv6 in brackets as invalid hostname format.

PR #276, opened 2015-10-06
Add IPv6 Host support to SSH.

jbcode42:
Changes REGEX to allow for IPV6 entry
Alters Uri generation to support IPV6 hosts on create


jbcode42:
I successfully tested this using an IPV6 IP Address and checked with backwards compatibility using IPV4 based hostnames. The one catch is the error when your device does not have an IPV6 address is a tad confusing. It simply says "connect failed: ENETUNREACH (Network is unreachable)" which is quite true if you do not have an IPV6 address or internet set up correctly, but may take a second to figure out.

kruton:
Did you test this on older devices? I don't think IPv6 parsing in the Uri class works for pre-KitKat or something.

jbcode42:
Not yet. I can do so in the next 24 hours. How far back do we hope to support?  If not, would adding a version check to determine the regex be acceptable?

khorimoto:
We need to support back to API v8.

kruton:
I started working on this a while ago, but didn't finish it. It might give you some hints as to what's wrong:
https://github.com/kruton/connectbot/tree/transportAddress

rhansby:
@jbcode42, will this solve #36?

jbcode42:
@rhansby Yes, however I haven't checked backwards compatibility yet, but it will allow it to take IPV6 literals. I was able to successfully input the IPV6 address in #36 and attempt to connect. It failed because I was not on an IPV6 network at the time (And even then, I don't know if it is a valid host test), but it did recognize the full IPV6 address and correctly recognize it as a hostname and at least attempt to connect to it.

jbcode42:
Tested fix on HTC Inc running CM 7.1 (Gingerbread 2.3.7 API Level 10) and it worked. Attempting to boot a emulator failed, so instead I elected to downgrade the HTC Inc again to CM 6.1 (Froyo 2.2.1 API Level 8). Confirmed working.

I think the reason this works is how the parseHost() call works in android.net.Uri. Normally it would split on the ":" symbol unless we encode the hostname. I looked at the source code from 2.3.2_r1 on GrepCode and it shows the following:

```
        private String parseHost() {
            String authority = getEncodedAuthority();
            if (authority == null) {
                return null;
            }

            // Parse out user info and then port.
            int userInfoSeparator = authority.indexOf('@');
            int portSeparator = authority.indexOf(':', userInfoSeparator);

            String encodedHost = portSeparator == NOT_FOUND
                    ? authority.substring(userInfoSeparator + 1)
                    : authority.substring(userInfoSeparator + 1, portSeparator);

            return decode(encodedHost);
        }
```

I do show that android version 2.0 does not bother with encoding/decoding the host string so my fix would not work on that version. GrepCode does not have an intermediate version available for reading and since it worked in my test, I have not tried to pull the relevant code from git.

jbcode42:
Is there anything else needed on this?

kruton:
I haven't had time to test this yet. Do shortcuts on the desktop still work?

jbcode42:
Yes, tested on both the older device and the newer device previously
mentioned. Just checking in because I will likely be unable to code or test
anything further for the next couple days.

kruton:
It should be good, then. I'll test it in the next day or so.

jbcode42:
Cool. I hope sometime next week when I have more time to see about Telnet IPV6. Thanks!

kruton:
Fixes #36

Issue #265, opened 2015-09-25
Long tapping the terminal only offers Paste, but should offer Copy and Paste


wp9015362:
Hello,

when long tapping (pressing and holding) on the terminal, the user is only being offered with "Copy". Once the user has copied something, long pressing on the terminal only offers "Paste".

But it should offer "Copy" ***and*** "Paste", otherwise it would only be possible to paste the previously copied text and would no longer be possible to copy anything else (at least not without having to resort to the toolbar menu).

So, could you please fix this and make long pressing on the terminal also offer "Copy" ***and*** "Paste" after something has been copied?

Regards

jklein24:
> when long tapping (pressing and holding) on the terminal, the user is only being offered with "Copy". Once the user has copied something, long pressing on the terminal only offers "Paste".

I'm pretty sure it *only* ever shows paste at the moment and never copy. For now, the user can use the action bar copy and paste buttons, but we're also looking into reworking the copy/paste flow altogether so  I'd rather not mess with this until then.

wp9015362:
> @jklein24 wrote:
> 
> I'm pretty sure it only ever shows paste at the moment and never copy. For now, the user can use the action bar copy and paste buttons,

Yes, sorry, you are correct, it actually only ever offers paste.

> @jklein24 wrote:
> 
> For now, the user can use the action bar copy and paste buttons

Yeah, but in https://github.com/connectbot/connectbot/issues/263 @alescdb mentioned that some users might have the toolbar hidden, which actually lead me to open this issue.

> @jklein24 wrote:
> 
>  but we're also looking into reworking the copy/paste flow altogether so I'd rather not mess with this until then.

May I ask what that will look like?

And is there some kind of to-do list for ConnectBot which could be made publicly available maybe?

Regards

wp9015362:
PS:

> @jklein24 wrote:
> 
>  but we're also looking into reworking the copy/paste flow altogether so I'd rather not mess with this until then.

Ah, I just saw that you are probably referring to this (?):

https://github.com/connectbot/connectbot/pull/129

Regards

jklein24:
Yep, exactly :-). #129 is roughly the goal.

rhansby:
This issue should be closed, it was fixed by #277

PR #277, opened 2015-10-06
TextView in terminal for better text selection

### LINKED ISSUE #232: Mouse-based selection of text
Click-dragging on text with a mouse should give the user a way to select that text. Jeremy, I remember you had thoughts on the exact way we wanted to implement this.

### LINKED ISSUE #270: Move page-up/page-down gestures to opt-in in settings
We have virtual keys for pg up and pg down; can we remove the "drag on the left side of the screen" gesture? @kruton @jklein24

rhansby:
Fixes #232, #265, and possibly other issues.

Accidentally fixes #270. Should it? 

Extends off of #129. Big thanks to @alescdb for the initial work! (I recommend that #129 be closed now)

Long pressing on text now allows users to naturally select text as they would elsewhere in the OS. Font changes are also handled properly (see GIF).

Pre-Honeycomb devices will use the old selection method, as text selection APIs are not available for them.

GIF attached:

output-opt

### DIFF
--- a/app/src/main/java/org/connectbot/ConsoleActivity.java
+++ b/app/src/main/java/org/connectbot/ConsoleActivity.java
@@ -30,7 +30,6 @@ import org.connectbot.service.TerminalKeyListener;
 import org.connectbot.service.TerminalManager;
 import org.connectbot.util.PreferenceConstants;
 
-import android.annotation.TargetApi;
 import android.app.AlertDialog;
 import android.app.Dialog;
 import android.content.ComponentName;
@@ -662,6 +661,15 @@ public class ConsoleActivity extends AppCompatActivity implements BridgeDisconne
 		if (tabs != null)
 			setupTabLayoutWithViewPager();
 
+		pager.setOnClickListener(new OnClickListener() {
+			@Override
+			public void onClick(View v) {
+				if (keyboardGroup.getVisibility() == View.GONE) {
+					showEmulatedKeys(false);
+				}
+			}
+		});
+
 		if (Build.VERSION.SDK_INT < Build.VERSION_CODES.HONEYCOMB) {
 			pager.setOnTouchListener(new OnTouchListener() {
 				public boolean onTouch(View v, MotionEvent event) {
@@ -669,10 +677,6 @@ public class ConsoleActivity extends AppCompatActivity implements BridgeDisconne
 
 					boolean isCopyingInProgress =
 						(copySource != null && copySource.isSelectingForCopy());
-
-					if (!isCopyingInProgress && keyboardGroup.getVisibility() == View.GONE) {
-						showEmulatedKeys(true);
-					}
 
 					// when copying, highlight the area
 					if (isCopyingInProgress) {
--- a/app/src/main/java/org/connectbot/ConsoleActivity.java
+++ b/app/src/main/java/org/connectbot/ConsoleActivity.java
@@ -137,9 +137,6 @@ public class ConsoleActivity extends AppCompatActivity implements BridgeDisconne
 
 	private MenuItem disconnect, copy, paste, portForward, resize, urlscan;
 
-	protected TerminalBridge copySource = null;
-	private int lastTouchRow, lastTouchCol;
-
 	private boolean forcedOrientation;
 
 	private Handler handler = new Handler();
@@ -669,78 +666,6 @@ public class ConsoleActivity extends AppCompatActivity implements BridgeDisconne
 				}
 			}
 		});
-
-		if (Build.VERSION.SDK_INT < Build.VERSION_CODES.HONEYCOMB) {
-			pager.setOnTouchListener(new OnTouchListener() {
-				public boolean onTouch(View v, MotionEvent event) {
-					TerminalBridge bridge = adapter.getCurrentTerminalView().bridge;
-
-					boolean isCopyingInProgress =
-						(copySource != null && copySource.isSelectingForCopy());
-
-					// when copying, highlight the area
-					if (isCopyingInProgress) {
-						SelectionArea area = copySource.getSelectionArea();
-						int row = (int) Math.floor(event.getY() / bridge.charHeight);
-						int col = (int) Math.floor(event.getX() / bridge.charWidth);
-
-						switch (event.getAction()) {
-						case MotionEvent.ACTION_DOWN:
-							// recording starting area
-							if (area.isSelectingOrigin()) {
-								area.setRow(row);
-								area.setColumn(col);
-								lastTouchRow = row;
-								lastTouchCol = col;
-								copySource.redraw();
-							}
-							return true;
-						case MotionEvent.ACTION_MOVE:
-							/* ignore when user hasn't moved since last time so
-							 * we can fine-tune with directional pad
-							 */
-							if (row == lastTouchRow && col == lastTouchCol)
-								return true;
-
-							// if the user moves, start the selection for other corner
-							area.finishSelectingOrigin();
-
-							// update selected area
-							area.setRow(row);
-							area.setColumn(col);
-							lastTouchRow = row;
-							lastTouchCol = col;
-							copySource.redraw();
-							return true;
-						case MotionEvent.ACTION_UP:
-							/* If they didn't move their finger, maybe they meant to
-							 * select the rest of the text with the directional pad.
-							 */
-							if (area.getLeft() == area.getRight() &&
-									area.getTop() == area.getBottom()) {
-								return true;
-							}
-
-							// copy selected area to clipboard
-							String copiedText = area.copyFrom(copySource.buffer);
-
-							clipboard.setText(copiedText);
-							Toast.makeText(ConsoleActivity.this, getString(R.string.console_copy_done, copiedText.length()), Toast.LENGTH_LONG).show();
-							// fall through to clear state
-
-						case MotionEvent.ACTION_CANCEL:
-							// make sure we clear any highlighted area
-							area.reset();
-							copySource.setSelectingForCopy(false);
-							copySource.redraw();
-							return true;
-						}
-					}
-
-					return true;
-				}
-			});
-		}
 	}
 
 	/**
@@ -843,7 +768,7 @@ public class ConsoleActivity extends AppCompatActivity implements BridgeDisconne
 			copy.setEnabled(activeTerminal);
 			copy.setOnMenuItemClickListener(new OnMenuItemClickListener() {
 				public boolean onMenuItemClick(MenuItem item) {
-					startCopyMode();
+					adapter.getCurrentTerminalView().startPreHoneycombCopyMode();
 					Toast.makeText(ConsoleActivity.this, getString(R.string.console_copy_start), Toast.LENGTH_LONG).show();
 					return true;
 				}
@@ -1108,24 +1033,6 @@ public class ConsoleActivity extends AppCompatActivity implements BridgeDisconne
 		}
 
 		super.onSaveInstanceState(savedInstanceState);
-	}
-
-	/**
-	 * Only intended for pre-Honeycomb devices.
-	 */
-	private void startCopyMode() {
-		// mark as copying and reset any previous bounds
-		TerminalView terminalView = (TerminalView) adapter.getCurrentTerminalView();
-		copySource = terminalView.bridge;
-
-		SelectionArea area = copySource.getSelectionArea();
-		area.reset();
-		area.setBounds(copySource.buffer.getColumns(), copySource.buffer.getRows());
-
-		copySource.setSelectingForCopy(true);
-
-		// Make sure we show the initial selection
-		copySource.redraw();
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/connectbot/util/TerminalViewPager.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+package org.connectbot.util;
+
+import android.content.Context;
+import android.support.v4.view.ViewPager;
+import android.util.AttributeSet;
+import android.view.MotionEvent;
+
+public class TerminalViewPager extends ViewPager {
```
jklein24: Can you add a brief javadoc to this class to explain why it's needed.

jklein24:
This is awesome! Thanks for digging through the internals of TextView to make this work, @rhansby! I think we should have more than one reviewer on this since it's a pretty big change. It'll be hard to break it up by file since most of this is moving from one file to another. 

cc: @khorimoto @kruton

jklein24:
also fyi: @bengolds

app/src/main/java/org/connectbot/TerminalView.java
```diff
@@ -150,6 +170,282 @@ public TerminalView(Context context, TerminalBridge bridge) {
 
 		// Enable accessibility features if a screen reader is active.
 		new AccessibilityStateTester().execute((Void) null);
+
+		clipboard = (ClipboardManager) context.getSystemService(Context.CLIPBOARD_SERVICE);
+
+		setTextColor(0x00000000);
```
jklein24: Does "Color.TRANSPARENT" work here?

app/src/main/java/org/connectbot/TerminalView.java
```diff
@@ -150,6 +170,282 @@ public TerminalView(Context context, TerminalBridge bridge) {
...
+					int base = bridge.buffer.getWindowBase();
+					bridge.buffer.setWindowBase(base - Math.round(yDistance));
+					return true;
+				}
+			}
+		}
+		return super.onGenericMotionEvent(event);
+	}
+
+	private void copyBufferToText() {
```
jklein24: small nit: The word "copy" threw me off for a minute here because I was looking for text copying. I'd prefer "refreshTextFromBuffer" or something.

app/src/main/java/org/connectbot/TerminalView.java
```diff
@@ -150,6 +170,282 @@ public TerminalView(Context context, TerminalBridge bridge) {
...
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.HONEYCOMB &&
+			selectionActionMode != null) {
+			selectionActionMode.finish();
+			selectionActionMode = null;
+		}
+	}
+
+	@Override
+	protected void onSelectionChanged(int selStart, int selEnd) {
+		if (selStart <= selEnd) {
```
jklein24: Why is this bit needed?

app/src/main/java/org/connectbot/service/TerminalBridge.java
```diff
@@ -71,6 +71,7 @@
 	public int defaultBg = HostDatabase.DEFAULT_BG_COLOR;
 
 	protected final TerminalManager manager;
+	public TerminalView terminalView;
```
jklein24: Is this any different from the "parent" member on line 90?

app/src/main/java/org/connectbot/util/TerminalViewPager.java
rhansby: Done.

app/src/main/java/org/connectbot/TerminalView.java
rhansby: Done.

app/src/main/java/org/connectbot/service/TerminalBridge.java
rhansby: Good catch. It was redundant. Fixed.

app/src/main/java/org/connectbot/TerminalView.java
rhansby: Are you referring to 

    if (selStart <= selEnd) {

? Sorry, it's unclear which line you're referring to.

app/src/main/java/org/connectbot/TerminalView.java
rhansby: Thanks, I didn't like the name much either.

app/src/main/java/org/connectbot/TerminalView.java
jklein24: The whole override I suppose. It seems like this is something that Textview should really already handle. I'm curious what was wrong.

app/src/main/java/org/connectbot/TerminalView.java
```diff
@@ -150,6 +171,282 @@ public TerminalView(Context context, TerminalBridge bridge) {
...
+
+					totalY += distanceY;
+					final int moved = (int) (totalY / bridge.charHeight);
+
+					if (moved != 0) {
+						int base = bridge.buffer.getWindowBase();
+						bridge.buffer.setWindowBase(base + moved);
+						totalY = 0;
+
+						refreshTextFromBuffer();
```
jklein24: Doesn't this only really need to be done here if we're already in selection mode? Is there any noticeable performance impact here when scrolling while text is selected?

app/src/main/java/org/connectbot/TerminalView.java
rhansby: TerminalView maintains a currentSelection, which is used when the user presses Ctrl+Shift+V on hard keyboard. This is why onSelectionChanged is overridden.

Unfortunately sometimes quickly moving the mouse selection results in `selEnd` being less than `selStart`, crashing the app without this check when it his `substring()`.

jklein24:
LGTM, but going to wait for feedback from @kruton

app/src/main/java/org/connectbot/TerminalView.java
rhansby: Which line are you referring to?

jklein24: Yeah, 205. `refreshTestFromBuffer()`

app/src/main/java/org/connectbot/ConsoleActivity.java
```diff
@@ -669,255 +661,86 @@ public boolean onTouch(View v, MotionEvent event) {
...
-				return false;
 			}
-
-
 		});
 
-		pager.setLongClickable(true);
-		pager.setOnTouchListener(new OnTouchListener() {
+		if (Build.VERSION.SDK_INT < Build.VERSION_CODES.HONEYCOMB) {
+			pager.setOnTouchListener(new OnTouchListener() {
```
khorimoto: Would it be possible to add this listener in TerminalView instead? Then we would have both the pre- and post-Honeycomb code in the same spot. It's confusing to have the two different copy implementations spread throughout different files.

app/src/main/java/org/connectbot/TerminalView.java
```diff
@@ -25,47 +25,65 @@
...
 import android.view.inputmethod.EditorInfo;
 import android.view.inputmethod.InputConnection;
+import android.widget.TextView;
 import android.widget.Toast;
 import de.mud.terminal.VDUBuffer;
 import de.mud.terminal.vt320;
 
 /**
- * User interface {@link View} for showing a TerminalBridge in an
+ * User interface {@link TextView} for showing a TerminalBridge in an
```
khorimoto: Please add some high-level explanation of how this works (i.e., it reads in a bitmap and prints transparent text to provide selection).

app/src/main/java/org/connectbot/TerminalView.java
khorimoto: Shouldn't there be a performance issue when text is NOT selected? This means we're redrawing the invisible text when nothing is selected, so the user wouldn't notice it anyway...right?

app/src/main/java/org/connectbot/TerminalView.java
```diff
@@ -150,6 +171,282 @@ public TerminalView(Context context, TerminalBridge bridge) {
...
+	@TargetApi(11)
+	private void initSelectionCallback() {
+		this.setCustomSelectionActionModeCallback(new ActionMode.Callback() {
+			private static final int PASTE = 0;
+
+			@Override
+			public boolean onCreateActionMode(ActionMode mode, Menu menu) {
+				TerminalView.this.selectionActionMode = mode;
+
+				menu.add(0, PASTE, 2, "Paste")
```
khorimoto: "Paste" needs to be localized.

app/src/main/java/org/connectbot/TerminalView.java
```diff
@@ -150,6 +171,282 @@ public TerminalView(Context context, TerminalBridge bridge) {
...
+	}
+
+	@Override
+	public boolean performLongClick() {
+		refreshTextFromBuffer();
+		return super.performLongClick();
+	}
+
+	@Override
+	public boolean onTouchEvent(MotionEvent event) {
```
khorimoto: Please comment/explain this function and why it's doing different things based on the version.

app/src/main/java/org/connectbot/TerminalView.java
```diff
@@ -150,6 +171,282 @@ public TerminalView(Context context, TerminalBridge bridge) {
...
+	 * @param bridge
+	 * @return True if the event is handled.
+	 */
+	@TargetApi(14)
+	private boolean onMouseEvent(MotionEvent event, TerminalBridge bridge) {
+		int row = (int) Math.floor(event.getY() / bridge.charHeight);
+		int col = (int) Math.floor(event.getX() / bridge.charWidth);
+		int meta = event.getMetaState();
+		boolean shiftOn = (meta & KeyEvent.META_SHIFT_ON) != 0;
+		boolean mouseReport = ((vt320) bridge.buffer).isMouseReportEnabled();
```
khorimoto: bridge.buffer is cast to vt320 all over this file. Should we just create a typed instance variable for it instead?

app/src/main/java/org/connectbot/TerminalView.java
alescdb: Also, I had problems with performClick() on M preview in my tests. I've just try on my Nexus 9 Marshmallow (release) and I can't select anything, so it seems the problem is still here.

I resolved it by using a TouchListener on TerminalView (cf https://github.com/alescdb/connectbot/commit/cfbbf3d51e7292d0da1c9fafe41d9948599244ad#diff-431cd3b89538f02edadfe658476256b6R161)

Nice work btw :-)

app/src/main/java/org/connectbot/ConsoleActivity.java
rhansby: Agreed. I'll move it over.

app/src/main/java/org/connectbot/TerminalView.java
rhansby: @alescdb, did you experience that issue on an emulator or a real device?

app/src/main/java/org/connectbot/TerminalView.java
rhansby: @jklein24 please advise.

app/src/main/java/org/connectbot/TerminalView.java
alescdb: On a real Nexus 9 with 6.0 (MRA58K). I haven't tested it on my real Nexus 5 yet...

app/src/main/java/org/connectbot/TerminalView.java
jklein24: Hmm I guess there are a few options. What Kyle mentions is probably the easiest since the bridge never changes in TerminalView. It's a bit risky because if anyone ever adds code to change the bridge or change the buffer inside the bridge, this instance variable will also need to be updated. Another alternatives would be to either add the mouse functions to the VDUBuffer (not sure they really belong there) or switch the type of TerminalBridge.buffer to a vt320, which exposes a much bigger interface, but we're working around the interface anyway here...

app/src/main/java/org/connectbot/TerminalView.java
rhansby: Good catch. Fixed: 9b8794d

app/src/main/java/org/connectbot/TerminalView.java
rhansby: Done:  17f70f5

app/src/main/java/org/connectbot/TerminalView.java
rhansby: Accidentally solved this via: 3c8cb9a

app/src/main/java/org/connectbot/TerminalView.java
rhansby: @khorimoto: comments added: 08ff54fbea83095f7c015df79f0f0176ea028d4c

@alescdb: did you test this specific PR, or were you referring to your old PR? I'm having no issues on either M or any other API version.

app/src/main/java/org/connectbot/TerminalView.java
rhansby: Fixed: 35f62aa

Went with Kyle's suggestion.

kruton:
Looks good to me.

PR #279, opened 2015-10-07
Make the pg up/down gesture a preference which is disabled by default.

### LINKED ISSUE #270: Move page-up/page-down gestures to opt-in in settings

jklein24:
Fixes #270


jklein24:
OK, this should finally be good to go :-)

PR #280, opened 2015-10-07
Partial implementation of DCESAVE for mouse reporting

jklein24:


app/src/main/java/de/mud/terminal/vt320.java
```diff
@@ -2197,10 +2198,20 @@ else if (R >= height)
...
+              case 1000:
+              case 1001:
+              case 1002:
+              case 1003:
+                mouserptSaved = mouserpt;
+                break;
+              default:
+                debug("ESC [ ? " + DCEvars[0] + " s, unimplemented!");
+              }
+            }
```
kruton: break;

jklein24: done.

PR #283, opened 2015-10-09
Update checkstyle and fix violations

kruton:


kruton:
This should fix the bug you ran into, @khorimoto

khorimoto:
Thanks :)

PR #289, opened 2015-10-13
Added an Inputfield so keyboard predictions can be used

AshyIsMe:
Based on the feature in irssi-connectbot.
I've added a button to the virtual keyboard that enables the password Inputfield in plaintext mode to allow keyboard predictions to be used.
To better fit this on the virtual keyboard I've split it in to two rows with just the Inputfield button and the show keyboard button on the top row.

Let me know if there's anything about this you'd prefer me to change before merging.


trygveaa:
Nice! I'd really like this feature.

@AshyIsMe: I made some minor changes/cleanup, fixed the conflicts against master and pushed it to https://github.com/trygveaa/connectbot/tree/inputfield-modified. Feel free to take a look at it and include it in the PR if you like.

Also, I found two issues with this when having the "Special keys always visible" setting activated.

1. After you type something in this input field and press enter, the keyboard row does not become visible again. It appears when you press the terminal, but I think it should appear when the input field disappears.
2. When this setting is active, the terminal ends above the keyboard row, instead of appearing beneath it (this is not a change in this PR). Since you have made the keyboard row two rows high, there is a lot of empty space to the left of the two buttons. I think you should consider changing it back to all the buttons on one row.

Issue #287, opened 2015-10-12
Scrolling causes selection to start during sustained touch


kruton:
If you use your finger to scroll the screen up and down eventually (after a touch timeout, it appears) selection will start no matter if your finger has stopped moving or not. The intended selection behavior is that a scroll will cancel the selection timer and just allow scrolling.

PR #290, opened 2015-10-13
Correct scrolling of selected text in TerminalView

### LINKED ISSUE #286: Selection stays active while text is scrolling
If you select a text and then type 'ls' the selection stays active and at the same place.

rhansby:
Fixes #286 and #287

GIFs included below:

Before:
before

After:
after


app/src/main/java/org/connectbot/TerminalView.java
```diff
@@ -250,55 +237,18 @@ public boolean onSingleTapConfirmed(MotionEvent e) {
...
-		}
-	}
-
-	@Override
-	protected void onSelectionChanged(int selStart, int selEnd) {
-		if (selStart <= selEnd) {
-			currentSelection = getText().toString().substring(selStart, selEnd);
-		}
-		super.onSelectionChanged(selStart, selEnd);
+		terminalTextViewOverlay.copyCurrentSelectionToClipboard();
```
kruton: This could be `null`, but is this even used anymore?

rhansby: Added a null check. And yes it's still used when the user presses `Ctrl`+`Shift`+`C`

jklein24:
LGTM, but would definitely feel more comfortable with @kruton's approval.

PR #293, opened 2015-10-13
Fix paste button being permanently disabled when clipboard was previously empty.

### LINKED ISSUE #292: When ActionBar is set to auto hide hitting paste icon is impossible
As sayed in #285
I can hit it when not set to auto hide.

rhansby:
Fixes #292.


jklein24:
So this will make the button enabled even if nothing's on the clipboard, right? I guess that just means pressing it is a no-op, which isn't terrible.

rhansby:
This doesn't change previous behavior -- before, the disabled paste button was still visible, but did not indicate it was disabled. Now, the enabled button will "ripple" when touched, at least indicating that the button works.

PR #294, opened 2015-10-13
Change PubkeyListActivity's options menu.

### LINKED ISSUE #168: Please use floating action button for pubkeys
Hello,

I am not a fan of the current UI for adding hosts.

But I have a suggestion to improve it:

The Google Material Design offers a UI element called "Floating Action Button".

You can read about it in the Google Material Design guidelines over there:

https://www.google.com/design/spec/components/buttons-floating-action-button.html

Could you please use a floating action button for adding new hosts?

The button should be placed in the lower right corner of the hosts screen and should make use of the ["ic_add" Material icon](https://www.google.com/design/icons/#ic_add) and when being pressed it should open a new dialog which would offer fields for protocol, hostname (IP), port and username.

Of course the current UI for adding new hosts should then be removed from the hosts screen.

Any chance you could do that? It would make for a much better UI / UX IMHO.

Regards

khorimoto:
Now, it has two icons to represent the two actions, as recommended by @bengolds. Fixes #168.


jklein24:
Looks like you forgot to git add some files

khorimoto:
Added the files. Oops.

PR #296, opened 2015-10-14
Fix paste button being permanently disabled because SSH session was not yet open.

### LINKED ISSUE #295: Can't hit Paste (SSH only)
OK I think I nailed it.
It only happens with SSH host. When connecting and after it is impossible to hit paste, if I select "Force size" in the menu, it then enable the Paste icon.

Nexus 5 with 6.0.

rhansby:
Fixes #295.

I'd prefer this not be merged until @alescdb weighs in on it.


rhansby:
Accidentally included a useless new file, so closed.

PR #303, opened 2015-10-20
Set more padding on volume pref SeekBar for Lollipop and above

### LINKED ISSUE #210: Bell volume adjustment slider is too wide
See https://www.google.com/design/spec/components/sliders.html# for better ways to do the slider. 

See attached screenshot.

bell volume awkward

johnheroy:
Improves the visual appearance of the VolumePreference Seekbar for Lollipop and above. See pics for below for screenshots across APIs 10, 15, and 23 post this change. Fixes #210.

# API v23
api v23

# API v15
api v15

# API v10
api v10


app/src/main/java/org/connectbot/util/VolumePreference.java
```diff
@@ -56,7 +57,13 @@ protected View onCreateDialogView() {
 		sb.setMax(100);
 		sb.setProgress((int) (getPersistedFloat(
 				PreferenceConstants.DEFAULT_BELL_VOLUME) * 100));
-		sb.setPadding(10, 10, 10, 10);
+
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+			sb.setPadding(75, 70, 75, 10);
```
jklein24: Did this come from the material spec somewhere?

johnheroy: No this did not come from the material spec, was just adjusting the padding so that the Seekbar has even padding and not hugging the top edge of the dialog.

Looking again at the [material spec](https://www.google.com/design/spec/components/sliders.html#sliders-continuous-slider) though we could do better by adding a volume icon to the left of the Seekbar. I could add this in another commit and add to the PR if you would like that change :)

kruton: Could this be moved to a resource? I didn't realize this was in the java code. It'd be best to put it in an XML layout, but it looks like this doesn't lend itself to that. Instead you could just add a `<dimen>` resource.

In a resource file such as `res/values/dimens.xml` you could add:
```
<dimen name="volume_dialog_padding_horizontal">75dp</dimen>
<dimen name="volume_dialog_padding_vertical">10dp</dimen>
```

Then you could do:
```
int paddingHorizontal = (int) getResources().getDimension(R.dimen.volume_dialog_padding_horizontal);
int paddingVertical = (int) getResources().getDimension(R.dimen.volume_dialog_padding_vertical);
sb.setPadding(paddingHorizontal, paddingVertical, paddingHorizontal, paddingVertical);
```

jklein24:
The padding specs for a dialog are here:

https://www.google.com/design/spec/components/dialogs.html#dialogs-specs

It looks like 75px is roughly 24dp in your screenshot, but that might not be true on all devices. I'd make sure it's 24dp.

app/src/main/java/org/connectbot/util/VolumePreference.java
jklein24: Better yet:

getContext().getResources().getDimension(android.support.design.R.dimen.abc_select_dialog_padding_start_material)

and

getContext().getResources().getDimension(android.support.design.R.dimen.abc_dialog_list_padding_vertical_material)

johnheroy:
I added the volume icon to match the slider examples in the material docs and matched the appropriate spacing (24dp to left of icon, content 80dp from left of dialog). To make this easier I factored out the dialog into an XML layout file, screenshots for the various versions as below:

# API v23
api v23

# API v15
api v15

# API v10
api v10

johnheroy:
I'm not totally sure why the [Travis CI build](https://travis-ci.org/connectbot/connectbot/builds/86517038) is failing, I can just see that there was a lint error somewhere at my ImageView declaration but not sure what...is there a way I can run the linter locally and view the precise error?

In any case I have a hunch it could be because I'm missing the "android:contentDescription" attribute. Is the general practice to add a string first (something like "Bell volume") to the English strings.xml and then it would be translated later?

jklein24:
Unfortunately, there's not a great process with the linter. You'll nee to run `./gradlew app:lint` and then check out the output html. You should be able to track down the lint errors in the files you've changed here.

On another note, sorry to nit-pick but the space between the icon and slider seems too big to me for some reason. Is there extra padding there on accident or something?

kruton:
Yes, if you run `./gradlew lint` then look at `app/bplid/outputs/lint-results.html` you'll see the error is:
```
[Accessibility] Missing contentDescription attribute on image
```

alescdb:
Just a trick (hint?) : You may also start lint from Android Studio with menu Analyze -> Inspect Code, travis-ci build gives you the file and the line where the new lint warning/error is.
In your case : 

```
$ ./scripts/check-lint-count.sh app/build/outputs/lint-results.xml $HOME/.cache/lint/lint-results.xml
Historical count : 552, new count : 553
FAILURE: lint issues increased from 552 to 553
497a498
> <location file="/home/travis/build/connectbot/connectbot/app/src/main/res/layout/volume_preference_dialog_layout.xml" line="29" column="5"/>
```

johnheroy:
Thanks for the catch, it looked like there was some inner padding on the SeekBar in Lollipop itself so I just reset with android:padding="0dp" for that element and added a contentDescription for the ImageView for a11y. Updated screencaps below:

# API v23
api v23

# API v14
api v14

# API v10
api v10

app/src/main/java/org/connectbot/util/VolumePreference.java
```diff
@@ -46,20 +48,17 @@ public VolumePreference(Context context, AttributeSet attrs, int defStyle) {
...
+	protected void onBindDialogView(View view) {
+		SeekBar volumeBar = (SeekBar) view.findViewById(R.id.volume_bar);
+		volumeBar.setProgress((int) (getPersistedFloat(
 				PreferenceConstants.DEFAULT_BELL_VOLUME) * 100));
-		sb.setPadding(10, 10, 10, 10);
-		sb.setOnSeekBarChangeListener(this);
-
-		return sb;
+		volumeBar.setOnSeekBarChangeListener(this);
+		super.onBindDialogView(view);
```
jklein24: Should this be done first in the method?

johnheroy: I had seen a lot of examples calling the super class's method at the end but tested both ways and does not seem to have made any difference, for convention I moved to the first line of the method.

kruton:
I just updated the lint failure output so it's more clear what is causing the issue.

johnheroy:
FYI, commit 9bceeb in conjunction with these changes made the seekbar styling on API v10 look consistent with v14/15 and v23. See below:

# API v23
api v23

# API v15
api v15

# API v10
api v10

PR #304, opened 2015-10-20
Polish UI for add/edit host.

khorimoto:
This finalizes the add/edit host view, except for these:

1) Introduce "do you want to discard edits" dialog when user clicks back.
2) Include the correct icons instead of the ConnectBot icon for every icon.
3) Grey out the "save/edit" icon when the host is invalid.

### DIFF
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -16,7 +16,7 @@
  ~ limitations under the License.
  */
 -->
-<resources>
+<resources xmlns:tools="http://schemas.android.com/tools">
 
 	<style name="AppTheme" parent="Theme.AppCompat">
 		<item name="colorPrimary">@color/primary</item>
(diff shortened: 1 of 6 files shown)


app/src/main/res/values/styles.xml
```diff
@@ -36,14 +36,38 @@
 		<item name="android:background">@drawable/keyboard_button_selector</item>
 	</style>
 
+	<style name="ListItemIcon">
+		<item name="android:layout_marginLeft">16dp</item>
```
jklein24: Should this use `@dimen/abc_list_item_padding_horizontal_material`?

Check out https://github.com/android/platform_frameworks_support/blob/master/v7/appcompat/res/values/dimens_material.xml#L36

Maybe there's a theme or style to what you're doing here in the support library too.

app/src/main/res/values/styles.xml
```diff
@@ -36,14 +36,38 @@
...
 	<style name="ListItemFirstLineText">
+		<item name="android:layout_alignParentTop">true</item>
 		<item name="android:textColor">?android:textColorPrimary</item>
 		<item name="android:textSize">16sp</item>
+		<item name="android:paddingTop">20dp</item>
+		<item name="android:layout_alignParentLeft">true</item>
+	</style>
+
+	<style name="ListItemFirstLineText.WithIcon" parent="ListItemFirstLineText">
+		<item name="android:layout_marginLeft">72dp</item>
```
jklein24: I feel like maybe some of this stuff should be pulled out into constants in dimens.xml because we have a lot of repeated values.

app/src/main/res/values/styles.xml
khorimoto: Done.

app/src/main/res/values/styles.xml
khorimoto: Done.

jklein24:
I think you forgot to push your new changes

khorimoto:
Oops, you're right. Pushed.

app/src/main/res/values/styles.xml
```diff
@@ -36,14 +36,38 @@
...
+		<item name="android:layout_centerVertical">true</item>
+		<item name="android:layout_alignParentLeft">true</item>
+		<item name="android:layout_alignParentStart" tools:ignore="NewApi">true</item>
+	</style>
+
 	<style name="ListItemFirstLineText">
+		<item name="android:layout_alignParentTop">true</item>
 		<item name="android:textColor">?android:textColorPrimary</item>
-		<item name="android:textSize">16sp</item>
+		<item name="android:textSize">@dimen/list_item_left_right_padding</item>
```
jklein24: This is a bit weird...  Did you mean to use padding for text size? Note that you should use sp instead of dp for font sizes because it can scale with the user's os-level font size preferences.

kruton:
LGTM

app/src/main/res/values/styles.xml
khorimoto: Fixed that. I had copy-pasted it to the wrong spot. Thanks for catching.

PR #305, opened 2015-10-20
Add/edit host dialog: Add save dialog and adjust icons.

khorimoto:
If the host is invalid and the user clicks the back button, pop up a dialog which asks the user whether to discard the changes or continue editing.

Additionally, change the "save" icon from the plus to the save icon, and make the icon translucent when it cannot be saved (due to an invalid URI).


jklein24:
LGTM. Will merge when tests pass

PR #308, opened 2015-10-20
Style port forward and pubkey list like the host list.

### LINKED ISSUE #264: Font size in Manage Pubkeys list and Port forwards list is way too big
Hello,

the font size in the Manage Pubkeys list is way too big.

It should follow the specs from https://www.google.com/design/spec/components/lists.html.

It's the same issue as it was with the host list in https://github.com/connectbot/connectbot/issues/175.

@khorimoto or @rhansby :

Any chance you could fix that as well, just as it was fixed in https://github.com/connectbot/connectbot/pull/185 for the host list?

PS: It would be nice if you could also do https://github.com/connectbot/connectbot/issues/253 while being at it, since the icons in the Manage Pubkeys list are also affected by https://github.com/connectbot/connectbot/issues/253 .

Regards

rhansby:
Fixes #264 and is an extension of #300 

Pubkey list before:
pubkey before

Pubkey list after: 
after maybe2

Port forward list after:
fwd after maybe

### DIFF
--- a/app/src/main/res/layout/item_host.xml
+++ b/app/src/main/res/layout/item_host.xml
@@ -30,13 +30,9 @@
 		android:id="@android:id/icon"
 		android:layout_width="40dp"
 		android:layout_height="40dp"
-		android:layout_alignParentStart="true"
-		android:layout_alignParentLeft="true"
-		android:layout_centerVertical="true"
-		android:layout_marginStart="16dp"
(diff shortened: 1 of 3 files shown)


kruton:
LGTM

jklein24:
Let's reuse @khorimoto's new lists styles when https://github.com/connectbot/connectbot/pull/304 lands

PR #314, opened 2015-10-21
Translations import

### LINKED ISSUE #311: Remove all unused string resources
The recent activity on ConnectBot has left a lot of string resources that are unused imposing a burden on anyone that wants to translate ConnectBot into a new language. We should remove all the unused resources in `app/src/main/res/values/strings.xml` (the other languages will be automatically updated by translations imports).

kruton:
Fixes #311


kruton:
Say what?
```
Historical count: 554, new count: 152
```

PR #318, opened 2015-10-21
Clear focus on font size EditText when font size SeekBar progress changed

### LINKED ISSUE #315: Host creation: font size slider value does not update when focused
When the user is focused on the value of the font size, dragging the slider does not change the value. 

GIF included:
output

johnheroy:
Add 
```
android:focusable="true"
android:focusableInTouchMode="true"
```
to the HostEditorFragment LinearLayout so focus doesn't jump back to the quick connect EditText when clearing focus on the font size EditText. In order though that we preserve a auto-focused quick connect container after tapping '+', manually request focus in the fragment's onCreate method.

Fixes #315.


app/src/main/res/layout/fragment_host_editor.xml
```diff
@@ -25,6 +25,8 @@
 	>
 
 	<LinearLayout
+		android:focusable="true"
```
khorimoto: Why do we need these changes?

johnheroy: Without these changes, when we clear focus on the font size EditText, focus automatically returns to the first focusable element, which is the quick connect container. It's just kind of janky UX to be fiddling with the font size and then have focus jump to a different section in the fragment.

khorimoto: As discussed offline, LGTM, but please add comments explaining why these are necessary. Thanks!

johnheroy: Done.

PR #319, opened 2015-10-21
Add new icons to add/edit host view.

khorimoto:


kruton:
Sorry, the latest merge from @alescdb broke your keyboard reference.

khorimoto:
Fixed. Thanks for catching it.

jklein24:
Should we be using svgs like @alescdb described in https://github.com/connectbot/connectbot/pull/321

khorimoto:
Switched to SVGs.

PR #320, opened 2015-10-22
Change the lint script to use message instead of summary

kruton:
The message has a better explanation (i.e., specific to the instance
it's talking about) than the summary which is fairly generic.


alescdb:
Maybe a `grep -E "^[+|-]file="` on `diff` output can help avoiding confusion ?

Issue #161, opened 2015-08-30
Virtual keyboard is using Emojis to display left and right arrow keys


wp9015362:
Hello,

the virtual keyboard in ConnectBot is using the "arrow backward" (U+25C0) Emoji to display the arrow left key and is using the "arrow forward" (U+25B6) Emoji to display the arrow right key.

Here is a screenshot which shows the issue:

![Image](https://cloud.githubusercontent.com/assets/7558360/9568498/86222e0c-4f4b-11e5-9112-36f6c7f4aa71.png)

Tested on an LG G2 running the stock LG Android 5.0.2 Lollipop ROM using ConnectBot 1.8.6 from the Google Play Store.

Since this is not happening for the up and down arrow keys, I would assume that this is a bug?

Could you please fix it?

Regards

wp9015362:
PS:

I just found the following pull request after having submitted this issue:

https://github.com/connectbot/connectbot/pull/127

So, this actually seems to be intentional.

But it still doesn't explain why it's only using the Emojis on the left and right arrow keys.

Anyway, IMHO you should rather use the Google Material icons for the virtual keyboard, which would also fix this issue.

I've submitted a feature request for it over there:

https://github.com/connectbot/connectbot/issues/162

Regards

kruton:
Argh, these geometric shapes were well before Emoji were standardized. I'm not sure why they co-opted this to code point to be in the Emoji group.

wp9015362:
> @kruton wrote:
> 
> I'm not sure why they co-opted this to code point to be in the Emoji group.

It is explained in the following video presentation of [Mark Davis from Google](https://plus.google.com/+MarkDavis/about), see (explanation starts at 07:35 min):

https://www.youtube.com/watch?v=-n9ONNeACyw

:wink:

The video also mentions a solution to this issue (starting at 08:35 min).

The solution is called "Emoji Variation Selectors", see:

http://www.unicode.org/reports/tr51/index.html#Emoji_Modifiers_And_Variation_Selectors
and:
http://www.unicode.org/reports/tr51/index.html#def_emoji_variation_selector

:smiley:

However, even though this issue (https://github.com/connectbot/connectbot/issues/161) could be fixed with Emoji Variation Selectors, I think it should rather be fixed by using Material icons as requested in https://github.com/connectbot/connectbot/issues/162.

Regards

nindanaoto:
This seems to be already fixed in the main.

PR #321, opened 2015-10-22
Replace unicode keys by vector drawable

alescdb:
Note that this need Android Studio 1.4 and gradle android plugin 1.4 (for backward compatibility)
This resolves issue #161

screenshot_20151022-101216


rhansby:
Sorry if this is a dumb question, but why did you choose to represent icons as SVGs in XML instead of converting them to PNGs?

alescdb:
Vector drawable are supported natively since lollipop (>=21), so you don't have to care about resolution and/or pixelisation. Since Android Studio 1.4/gradle 1.4 we get backward compatibility, so gradle creates the PNGs for all dpi for you automatically at build time.
So instead of having 5 PNGs (mdpi, hdpi, ...) to work with, you only have 1 xml file (also if you build a project for >= 21, your APK will be significantly smaller).
Other advantage, all Material Design icons (from https://www.google.com/design/icons/) are avalable for import directly in Android Studio (right clik menu New -> Vector Drawable). 
In my opinion, we should now use only vector drawable.
Of course this is only my opinion :-)

jklein24:
Cool! Happy to merge as soon as the linter is happy.

alescdb:
Lint is happy with me :-)

From what I see in travis-ci logs, lint warnings are old ones.
I think there is something wrong with the check-lint script (or it's cache)

rhansby:
Thanks for the explanation @alescdb!

alescdb:
@jklein24 I've checked and rechecked, lint warnings are not related to the files I've modified.
Aren't lint checks gradle plugin dependent ?

kruton:
Yes, it will have new lint checks.  But the "string/button_key_tab" lint error is caused by this change. I can switch to newer gradle in a different change.

alescdb:
Thanks @kruton and sorry about the missed button_key_tab, but when I remove it, do I need to removed it in all translations ? If I don't I get a lint error.

alescdb:
@kruton This time I'm 100% sure none of the warnings are mine :-)
Now the question is how to update or delete `$HOME/.cache/lint/lint-results.xml` on tavis-ci ?

kruton:
I didn't clear out the PR.321 cache before restarting the build. I've  just done that.

PR #329, opened 2015-10-23
Remove unused add host dialog from HostListActivity.

jklein24:


jklein24:
sigh... @kruton any idea what's up with the linter here? I just removed some code...

alescdb:
I'm not kruton but I guess it's dia_add_host (unused)

alescdb:
In fact all the warning starting with a '+' so :
R.string.button_add (and all translations)
R.string.button_cancel (and all translations)
...

kruton:
Yes, there's just a lot more you can remove since you removed all the
references to them :)

Issue #159, opened 2015-08-29
Option for showing virtual keys permanently


jplitza:
I like the new feature of having virtual arrow keys very much! That much, in fact, the I would love to have the bar show up at all times, not only when touching. That way, it would only take one touch to use the keys and I could look where to touch while typing other stuff without first having to touch the console.

wp9015362:
+1

wp9015362:
When the virtual keyboard would be displayed permanently, it's transparency should also be disabled IMHO.

PR #330, opened 2015-10-23
Add an option to always show the virtual keyboard (special keys)

alescdb:
#159

If someone can check my two resources strings to see if it is correct english or something that looks like english but is not :-)

```
<string name="pref_alwaysvisible_title">Keyboard always visible</string>
<string name="pref_alwaysvisible_summary">Always show the special keys</string>
```


app/src/main/res/values/strings.xml
```diff
@@ -644,4 +643,6 @@
 	<!-- Text for button which does not discard changes and keeps editing. -->
 	<string name="discard_host_cancel_button">Keep editing</string>
 
+	<string name="pref_alwaysvisible_title">Keyboard always visible</string>
```
jklein24: nit: I'd rather just have this say "Special keys always visible." not sure we even need the description. with "keyboard" it's a little ambiguous whether you mean the soft keyboard or special keys.

alescdb:
@jklein24 I've made the changes according to your suggestions.

app/src/main/java/org/connectbot/ConsoleActivity.java
```diff
@@ -87,6 +74,20 @@
...
+import org.connectbot.service.TerminalBridge;
+import org.connectbot.service.TerminalKeyListener;
+import org.connectbot.service.TerminalManager;
+import org.connectbot.util.PreferenceConstants;
+import org.connectbot.util.TerminalViewPager;
+
+import java.lang.ref.WeakReference;
+import java.util.ArrayList;
+import java.util.List;
+
```
kruton: Any reason to rearrange these?

app/src/main/res/values/strings.xml
```diff
@@ -644,4 +643,5 @@
 	<!-- Text for button which does not discard changes and keeps editing. -->
 	<string name="discard_host_cancel_button">Keep editing</string>
 
+	<string name="pref_alwaysvisible_title">Special keys always visible</string>
```
kruton: You need to have a comment above to provide context for translators.

app/src/main/java/org/connectbot/ConsoleActivity.java
alescdb: Corrected, sorry about that.
Everytime I touch code and/or xml in ConnectBot I'm struggling with Android Studio and my default settings. Indent with spaces is systematically selected for me in **all** code style configs and imports are automatically rearrange and optimized.

PR #332, opened 2015-10-26
Fix colors of positive/negative buttons in new host cancel dialog

### LINKED ISSUE #317: Host creation cancellation dialog uses incorrect colors
bad color

johnheroy:
* Add AlertDialogTheme, which defines an accent color consistent with the app
* Use this AlertDialogTheme in the new host cancellation dialog
* Refactor other instances of framework AlertDialog to use android.support.v7.app.AlertDialog instead with AlertDialogTheme to have material design dialogs going back to v7

Fixes #317.

Note: we can't use the android:alertDialogTheme attribute in AppTheme because it only works back to API 11 (won't work going back to API 8).


app/src/main/res/values/styles.xml
```diff
@@ -24,6 +24,10 @@
 		<item name="colorAccent">@color/accent</item>
 	</style>
 
+	<style name="AlertDialogTheme" parent="Theme.AppCompat.Dialog.Alert">
```
khorimoto: Would it make sense to have AppTheme be the parent? Or do you need to inherit from the dialog theme?

Alternatively, would it make sense just to use R.style.AppTheme when you instantiate the AlertDialog.Builder objects?

johnheroy: I thought about these approaches but in either case if the app theme applies to the AlertDialog then the dialog takes up the full screen and doesn't even look like a dialog:

full screen dialog

khorimoto: Cool...LGTM as-is, then.

PR #333, opened 2015-10-26
Do not apply background tint on top of FAB image

### LINKED ISSUE #316: Host creation: faded FAB
If you begin to create a host, but then go back and discard changes, the FAB is faded and seems disabled (it is not disabled).

Screenshot:
faded

johnheroy:
* Add `android:tint="@android:color/white"` to the add host FAB so that the image's tinted state is also white (and thus doesn't look disabled when you click back). [android docs](http://developer.android.com/reference/android/widget/ImageView.html#attr_android:tint)
* Refactor 'add' icon in the FAB to use a vector instead of pngs

Fixes #316.


khorimoto:
LGTM

johnheroy:
Pushed another commit which should fix the linter, android:tintMode was only API 21+ so am using android:tint instead with a color value (white).

khorimoto:
Still LGTM! :+1:

Issue #176, opened 2015-08-31
Please add "Quit" button to main menu


wp9015362:
Hello,

ConnectBot has a main menu which offers the following options/buttons:

- Sort by color/name
- Manage Pubkeys
- Colors
- Settings
- Help

Could you please add another button/option called "Quit" or "Exit" to the main menu which disconnects from all hosts and closes ConnectBot?

It would be much appreciated.

Regards

kruton:
I think "Disconnect all" would be useful, but not a "Quit" button. That goes against Android UX guidelines.

PR #364, opened 2016-01-07
Add DisconnectAll to HostActivity's Options.

arthurhsu:
Fix https://github.com/connectbot/connectbot/issues/176

### DIFF
--- a/app/src/main/java/org/connectbot/HostListActivity.java
+++ b/app/src/main/java/org/connectbot/HostListActivity.java
@@ -17,17 +17,6 @@
 
 package org.connectbot;
 
-import java.util.List;
-
-import org.connectbot.bean.HostBean;
-import org.connectbot.data.HostStorage;
-import org.connectbot.service.OnHostStatusChangedListener;
-import org.connectbot.service.TerminalBridge;
-import org.connectbot.service.TerminalManager;
-import org.connectbot.transport.TransportFactory;
-import org.connectbot.util.HostDatabase;
-import org.connectbot.util.PreferenceConstants;
-
 import android.content.ComponentName;
 import android.content.Context;
 import android.content.DialogInterface;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/connectbot/HostListActivity.java
```diff
@@ -233,6 +235,7 @@ public boolean onPrepareOptionsMenu(Menu menu) {
 
 		sortcolor.setVisible(!sortedByColor);
 		sortlast.setVisible(sortedByColor);
+		disconnectall.setEnabled(bound.getBridges().size() > 0);
```
jklein24: Maybe this should be setVisible instead of setEnabled. wdyt?

app/src/main/java/org/connectbot/HostListActivity.java
```diff
@@ -273,6 +276,17 @@ public boolean onMenuItemClick(MenuItem item) {
 		colors.setIcon(android.R.drawable.ic_menu_slideshow);
 		colors.setIntent(new Intent(HostListActivity.this, ColorsActivity.class));
 
+		disconnectall = menu.add(R.string.list_menu_disconnect);
+		disconnectall.setIcon(android.R.drawable.ic_menu_delete);
+		final HostListActivity self = this;
+		disconnectall.setOnMenuItemClickListener(new OnMenuItemClickListener() {
+			@Override
+			public boolean onMenuItemClick(MenuItem menuItem) {
+				self.disconnectAll();
```
jklein24: You can just call `disconnectAll()`. No need for the self stuff. See the calls to `updateList` in the above handlers.

app/src/main/java/org/connectbot/HostListActivity.java
arthurhsu: I think disabling is a better idea for better discovery. Users will know there's a way to disconnect all when not connected to any session.

jklein24:
LGTM

PR #375, opened 2016-02-28
Make sure the action bar is always shown on click

trygveaa:
If the action bar is set to auto hide, it should always be shown when
clicking the terminal. Previously, this did not happen when the virtual
keyboard was visible.


kruton:
Could you rebase this change?

trygveaa:
@kruton: Done.

PR #404, opened 2016-05-27
Use vector drawables

kruton:


alescdb:
vectorDrawables.useSupportLibrary should not be set to true unless you change code & xml to use the new support library.
Setting vectorDrawables.useSupportLibrary to true will disable gradle PNG generation (for backwards compatibility).

kruton:
Whoops, I thought we had done that already. Reverted with #406.

PR #442, opened 2016-09-13
Enable pinch to change font size

### LINKED ISSUE #430: Add pinch-to-zoom
ConnectBot VX allows changing the font size dynamically by using a pinch gesture. This feature can be quite useful when changing display orientation or effective area when moving to/from a hardware keyboard.

ConnectBot VX implementation: https://github.com/vx/connectbot/commit/3f4dd22becd5e0cd033ba4e5b574c15acd241a6d

morckx:
Resolves #430

### DIFF
--- a/app/src/main/java/org/connectbot/TerminalView.java
+++ b/app/src/main/java/org/connectbot/TerminalView.java
@@ -51,6 +51,7 @@ import android.text.ClipboardManager;
 import android.view.GestureDetector;
 import android.view.KeyEvent;
 import android.view.MotionEvent;
+import android.view.ScaleGestureDetector;
 import android.view.View;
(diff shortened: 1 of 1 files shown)


morckx:
With the current modest implementation, I did not experience any more interactions than with using the volume keys. My only interaction for both is that the textview overlay sometimes resets the window position. The volume key approach, in addition, has some interactions with changing the volume :) Or do you mean that scale gestures are mistaken as highlighting? Not any more than in a WebView on my devices. The issue that terminal and textview overlay sometimes seem to be out of sync when trying to highlight, I do have independently of zooming.

Anyway, I'll probably upload another commit that enables more real, continuous pinch-to-zoom, directly applying the scale-factor to the font size.

PR #443, opened 2016-09-13
Let keyboard button toggle keyboard visibility - reliably

### LINKED ISSUE #376: Keyboard does not open when pressing keyboard button
When pressing the keyboard button on the virtual keyboard row, the soft keyboard does not open (most of the time, see next paragraph). The only thing that happens is that the virtual keyboard row hides.

If you open a new session, close the soft keyboard and press the keyboard button before the virtual keyboard row hides for the first time, the button actually works and the soft keyboard opens again. However if the virtual keyboard row has been hidden at least once, the button does not work.

This is with the current newest master (c9682d3). The error seems to have been introduced by 381c4b6.

morckx:
* use the button not only to show, but also to hide the keyboard
* do this more reliably, i.e. even when the focus is not on the
terminal (fixes #376)

### DIFF
--- a/app/src/androidTest/java/org/connectbot/StartupTest.java
+++ b/app/src/androidTest/java/org/connectbot/StartupTest.java
@@ -1,6 +1,7 @@
 package org.connectbot;
 
 import org.connectbot.util.HostDatabase;
+import org.connectbot.util.PreferenceConstants;
 import org.hamcrest.Matcher;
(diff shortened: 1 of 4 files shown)


kruton:
Do you think you could add a test for this?

morckx:
I've never written a test for UI interaction, but I can give it a try.

PR #444, opened 2016-10-04
Customizable emulated keys

morckx:
This is of course only one possible solution for customizable emulated keys. I like it very much because it doesn't require any additional preferences and probably because it's fun dragging the tiny keys around.

Adding a test for the dragging operation seems currently impossible, with manageable effort. Espresso  doesn't really support such a kind of drag and drop, yet.


kruton:
This sounds cool. Do you mind making a little video clip to post of it in action? You can use [Telecine](https://play.google.com/store/apps/details?id=com.jakewharton.telecine&hl=en).

morckx:
Here is a little video: [https://youtu.be/9E90U3MV5jc](https://youtu.be/9E90U3MV5jc)
Note the little shaking of the keys in the beginning to indicate that they are dragable.

kruton:
That's cool. I think it might be too easy to get into this mode currently. What do you think about moving the enabling of the edit into the action bar overflow menu?

morckx:
Haha, and I thought you were a big opponent of adding more options :)
In this case, I think, this is really not necessary. In the unlikely case that a button is moved accidentally and additionally dropped on an accepted place, it can just easily be moved back.
OTOH such on option in the overflow menu would give a hint that dragging is possible. What I'd rather need after using this since some time, would be a "reset special key order" option in the settings menu.

kruton:
Sorry, I accidentally merged something to master without going through the check. Fixing it now.

Issue #450, opened 2016-11-18
Telnet [TTYPE] TELNET Option negotiation doesn't respond. 


M-griffin:
Telnet [TTYPE] TELNET Option negotiation would be a nice feature where the user can select ANSI, XTERM to give the server an idea of what features and options are supported by the connect connectbot application.

M-griffin:
This an intermittent issues with NAWS detection are addressed in the following Pull Request #451

M-griffin:
Sweet, Thanks!

PR #451, opened 2016-11-19
Telnet Protocol Handler Fixes

M-griffin:
This corrects intermittent issues with servers not receiving complete TELOPT  sequence causing incorrect terminal and screen size detection.

Built Telnet Response Sequences in to an ArrayList<Byte> to transpose back to a byte[] array so all response sequences are sent in a single write() instead of sending each separate byte in an individual write().

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -64,8 +64,8 @@ android {
 
     defaultConfig {
         applicationId "org.connectbot"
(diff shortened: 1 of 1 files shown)


M-griffin:
This addresses Issue #450 
Thanks.

PR #454, opened 2016-11-22
Adds retry mechanism to reconnect for more robustness

### LINKED ISSUE #439: Reconnecting after short network connectivity loss
Hi
ConnectBot completely fails re-connecting when network connection drops and then reconnects back on again. This happens mostly when driving/walking through areas with bad coverage.

Observing the app when this happens I can see it tries to reconnect right after connectivity loss, this obviously makes it fail (no route to server), then, when network is on again, because it failed connecting once it does not try reconnecting again.

I use ConnectBot to remotely control my home machines, also, I make much use of SSH tunnels, but when ConnectBot breaks connection with server and it's not able to reconnect again, of course the tunnels stop working, any background app using them also stops working.One example: I use GPSLogger to log GPS data to my own server (personal tracking for emergency situations when I go out riding..), this stops working because it needs the SSH tunnel open to send coordinates, ConnectBot wasn't able to reconnect when network connectivity was good again.

It would be GREAT if ConnectBot could detect network connectivity state and wait for it to make the connection. I don't think it would be too much trouble to do this.

ConnectBot it's pretty much the only reliable open source (and even closed source, not much more SSH clients out there) SSH client for android, but I think it lacks many 'basic' features.

Thanks for your time.

fusiondog:
Not too fancy but implements a retry mechanism that falls back in frequency and eventually stops.  Should be effective to reconnect in most cases. Fixes #439

### DIFF
--- a/app/src/main/java/org/connectbot/service/TerminalManager.java
+++ b/app/src/main/java/org/connectbot/service/TerminalManager.java
@@ -721,7 +721,7 @@ public class TerminalManager extends Service implements BridgeDisconnectedListen
 					continue;
 				}
 
-				int sleepVal = 1500;
+				int sleepVal = 500;
 				int retryLimit = getRetries();
 				int retryCount = 0;
 
@@ -746,7 +746,16 @@ public class TerminalManager extends Service implements BridgeDisconnectedListen
 					}
 
 					retryCount++;
-					Log.d(TAG, String.format("Retries %s", retryCount));
+					Log.d(TAG, String.format("Retries %s session stat: %s", retryCount, bridge.isSessionOpen() ));
+
+					while(bridge.isConnecting()){
+						try {
+							Thread.sleep(500);
+						} catch(InterruptedException ex) {
+							Thread.currentThread().interrupt();
+						}
+						Log.d(TAG, String.format("Sleeping while wait on connecting"));
+					}
 				}
 			}
 			mPendingReconnect.clear();
--- a/app/src/main/java/org/connectbot/service/TerminalManager.java
+++ b/app/src/main/java/org/connectbot/service/TerminalManager.java
@@ -711,7 +711,23 @@ public class TerminalManager extends Service implements BridgeDisconnectedListen
 				if (bridge == null) {
 					continue;
 				}
-				bridge.startConnection();
+				int sleepVal = 1000;
+				int retryLimit = 0;
+				int retryCount = 50;
+				while ( ( retryLimit >= 0 || retryCount <= retryLimit ) && bridge.isDisconnected() ){
+					try {
+						Thread.sleep(sleepVal);
+					} catch(InterruptedException ex) {
+						Thread.currentThread().interrupt();
+					}
+
+					bridge.startConnection();
+					if ( retryCount < 8 ) {
+						sleepVal = sleepVal * 2;
+					}
+					retryCount++;
+					Log.d(TAG, String.format("Retries %s", retryCount));
+				}
 			}
 			mPendingReconnect.clear();
 		}
--- a/app/src/main/java/org/connectbot/service/TerminalManager.java
+++ b/app/src/main/java/org/connectbot/service/TerminalManager.java
@@ -120,6 +120,11 @@ public class TerminalManager extends Service implements BridgeDisconnectedListen
 
 	public boolean hardKeyboardHidden;
 
+	public final static int MINIMUM_RETRY_INTERVAL = 1000; // minimum milliseconds to delay waiting for network and SSH connections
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/connectbot/service/TerminalManager.java
```diff
@@ -711,7 +720,43 @@ private void reconnectPending() {
...
+				int retryLimit = getRetries();
+				int retryCount = 0;
+
+				try {
+					Thread.sleep(sleepVal);
+				} catch(InterruptedException ex) {
+					Thread.currentThread().interrupt();
+				}
+
+				while ( ( retryLimit == 0 || retryCount <= retryLimit ) && bridge.isDisconnected() ){
```
kruton: The spacing style isn't following the existing conventions. Can you reformat this patch to match what exists?

app/src/main/java/org/connectbot/transport/SSH.java
```diff
@@ -439,6 +440,7 @@ private void finishConnection() {
 
 	@Override
 	public void connect() {
+		connecting = true;
```
kruton: It might be easier to follow if you extract this out to another function like:

```
public void connect() {
    connecting = true;
    try {
        connectInternal();
    } finally {
        connecting = false;
    }
}

private void connectInternal() {
    // previous contents of connect() ...
}
```

app/src/main/res/values/strings.xml
```diff
@@ -265,6 +265,11 @@
 	<!-- Summary for the preference that forces the service to stay running in the background. -->
 	<string name="pref_conn_persist_summary">"Force connections to stay connected while in background"</string>
 
+	<!-- Name for the retries size preference -->
+	<string name="pref_retries_title">"Persist retries"</string>
+	<!-- Description of the retries size preference -->
```
kruton: A bit more context of where this string appears will help. Translators only see these comments and not the context of what is retrying, where in the UI this will be seen, or what it is used for.

app/src/main/java/org/connectbot/service/TerminalManager.java
```diff
@@ -711,7 +720,43 @@ private void reconnectPending() {
...
+
+					bridge.startConnection();
+
+					if ( retryCount < 8 ) {
```
kruton: Any kind of constants ("8" here and "2" below) should be extracted out to be in the class near the top.

app/src/main/java/org/connectbot/service/TerminalManager.java
```diff
@@ -711,7 +720,43 @@ private void reconnectPending() {
...
+					try {
+						Thread.sleep(sleepVal);
+					} catch(InterruptedException ex) {
+						Thread.currentThread().interrupt();
+					}
+
+					retryCount++;
+					Log.d(TAG, String.format("Retries %s session stat: %s", retryCount, bridge.isSessionOpen() ));
+
+					while(bridge.isConnecting()){
```
kruton: Hmm, this could go on forever and drain batteries. There should be some kind of limit for how long it will wait.

app/src/main/java/org/connectbot/service/TerminalManager.java
```diff
@@ -711,7 +720,43 @@ private void reconnectPending() {
 				if (bridge == null) {
 					continue;
 				}
-				bridge.startConnection();
+
+				int sleepVal = 500;
```
kruton: This should be extracted to a constant.

app/src/main/java/org/connectbot/service/TerminalManager.java
```diff
@@ -711,7 +720,43 @@ private void reconnectPending() {
...
+					} catch(InterruptedException ex) {
+						Thread.currentThread().interrupt();
+					}
+
+					retryCount++;
+					Log.d(TAG, String.format("Retries %s session stat: %s", retryCount, bridge.isSessionOpen() ));
+
+					while(bridge.isConnecting()){
+						try {
+							Thread.sleep(500);
```
kruton: Looks like you forgot to re-use `sleepVal` here.

app/src/main/java/org/connectbot/transport/SSH.java
fusiondog: This line didn't change with the addition of the intermediate function, but it is testing out ok on my build.

app/src/main/java/org/connectbot/service/TerminalManager.java
fusiondog: Using minimum retry interval here because it remains constant and this is to wait on the SSH connect which should be more of a known value and not exceed ~30 sec.  The sleepVal grows to address more general network issues which may not as quickly resolve.

kruton:
`NullTransport` is still broken in this PR.

PR #473, opened 2017-02-15
Add button for fdroid + change the one for google play

Poussinou:


kruton:
Unfortunately I do not control what f-droid ships, so I'm uncomfortable adding a link here.

Poussinou:
They simply build your app from sources, controlling there's no proprietary binaries or dependencies, you can maybe talk with them, but if you don't want this button I can close the PR

kruton:
How about no button, but text that it's also available on F-Droid with the caveat that it is not built by the ConnectBot team?

Poussinou:
As you want, I'll modify this tomorrow

Poussinou:
Is this ok now?

KamranMackey:
@Poussinou Looks good, but it could use a few typo fixes. `build` should be `built`, the a in Android should be capitalized, and the colon next to "find it here" should not have a space next to it.

Just my two cents.

Poussinou:
Sorry about that

PR #474, opened 2017-02-16
Update Google Play Services version

kruton:


kruton:
This requires a minApi level that is too high right now.

PR #539, opened 2017-07-08
Add fdroid badge, update google badge

mueller-ma:
and add legal attribution
https://play.google.com/intl/en_us/badges/


Poussinou:
#473 I'll close mine

Issue #521, opened 2017-04-12
URL Scan prefixes http:// incorrectly, causing browser confusion (suggested fix included)


tsumare:
In git-v1.9.2-50-g94c1907 (currently the latest available from Releases), when using URL Scan on my Nexus 6X on Android 7, the URL list shows my URLs correctly, but when I select one to launch it, it goes to chrome as "http//www..." without a : after the scheme.  It will keep the scheme itself: "http://" becomes "http//".  "https://" becomes "https//".

After a brief glance at the code, [these lines](https://github.com/connectbot/connectbot/blob/8cf548e559ab82ee9ee15533038b265094c8f733/app/src/main/java/org/connectbot/ConsoleActivity.java#L1188-L1189) may be at fault.  It appears that "http://" is added if there IS a scheme in the url already, rather than if there is NOT.

ilf:
I can confirm this bug. https URLs are opened as http://www.https.com//www.google.com/. #517 is another dupe. Would be cool to merge this pull request. Thanks!

Issue #517, opened 2017-04-04
URLs opened have http://"


donnm:

donnm:
Hit send by accident before I finished the report.

On invoking URL scan the opened link has "http://" prepended, breaking URLs in the destination app.

donnm:
Version is 1.9.2-27 installed from F-droid.

justdave:
#521 appears to be the same thing... Newer report but has a suggested fix

PR #550, opened 2017-10-01
do not prefix urls with http://

lotheac:
this was broken by inverted logic in
36b4b5ea92bc3ede9626d73503d9139f97044cc4 -- http:// was being prepended
to urls which already contained it. But since URL scan doesn't seem to
pick up URLs without scheme anyway, this check is redundant.

fixes https://github.com/connectbot/connectbot/issues/521 & https://github.com/connectbot/connectbot/issues/517


ilf:
Thanks! Since I run into this a lot, I would really love a release with this fix. "Release early, release often." :) Thanks, and keep up the good work!

donnm:
Any schedule on a release for this fix? I use the URL scan multiple times a day and am constantly editing "http://" out of the browser location. I first reported this bug in #517 almost a year ago.

PR #551, opened 2017-10-17
Fix the method to calculate scroll, fix #519

### LINKED ISSUE #519: Scrolling down is not as smooth as scrolling up
I'm using ConnectBot v1.9.2-50-g94c1907 on a Samsung Galaxy Note 4 with Android KK v4.4.4 and when scrolling long outputs (such as `ls -Rl` output) I noticed that the **text scrolls smoothly when going down** (i.e. when you keep your finger pressed, the text goes fast just by moving the finger a few pixels away).

Unfortunately the same is not true - or at least **it's not the same** - when scrolling **up** (i.e. if you keep the finger pressed you have to travel a longer distance to make the text scroll and it does not scroll continuosly but "in jerks").

screen

This is especially noticeable on long history texts: you go quickly up, but not so quickly down.
Here's a [video recorded while scrolling](https://www.dropbox.com/s/bf7pj15v61kvz6v/connectbot_scrolling.mp4?dl=1).

wangqr:
The original code [here](https://github.com/connectbot/connectbot/blob/94c1907df5a76726d3492e959c6c08778c893d76/app/src/main/java/org/connectbot/util/TerminalTextViewOverlay.java#L176) is setting the line position to int on every `scrollTo` event. So scrolling up several pixels will cause 1 line up, and scrolling down several pixels will cause nothing. This leads to #519.

In this PR, the round to int work is only done on touch up event.

There is another issue causing the scrolling no so smooth: [`refreshTextFromBuffer`](https://github.com/connectbot/connectbot/blob/94c1907df5a76726d3492e959c6c08778c893d76/app/src/main/java/org/connectbot/util/TerminalTextViewOverlay.java#L183) can run for seconds, so the view will not respond to scrolling for the first few seconds. I have no idea how to fix this. Is there any suggestions or I should open another issue?

### DIFF
--- a/app/src/main/java/org/connectbot/util/TerminalTextViewOverlay.java
+++ b/app/src/main/java/org/connectbot/util/TerminalTextViewOverlay.java
@@ -168,12 +168,12 @@ public class TerminalTextViewOverlay extends TextView {
 
 	@Override
 	public void scrollTo(int x, int y) {
-		int lineMultiple = y / getLineHeight();
+		int lineMultiple = (y * 2 + 1) / (getLineHeight() * 2);
 
 		TerminalBridge bridge = terminalView.bridge;
 		bridge.buffer.setWindowBase(lineMultiple);
 
-		super.scrollTo(0, lineMultiple * getLineHeight());
+		super.scrollTo(0, y);
 	}
 
 	@Override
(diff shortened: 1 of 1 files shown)


morckx:
Thanks for the fix! It works perfectly for me. @kruton :+1: please merge this.

Concerning the issues with refreshTextFromBuffer [https://github.com/morckx/connectbot/commit/4aa73c8ffea327e0685905bdd266cfec493b7aa5](https://github.com/morckx/connectbot/commit/4aa73c8ffea327e0685905bdd266cfec493b7aa5) might help. At least, this avoids unnecessary updates and makes the problem less likely to occur – possibly including related garbage collection issues.

Issue #533, opened 2017-06-03
Can't see cursor in Emacs


snan:
Just updated to the new version of connectbot and it's great but I can't see the cursor in Emacs anymore. I've got some liberal use of setq cursor-type and set-cursor-color, but it used to work fine.

I alternate between box and bar, black and red depending on which combination of modes I'm in. It used to be that on the old connectbot, they all showed up as just a black box, which was fine.

Now, they're all invisible.

(Being able to access the same session of Emacs whether I'm on my phone or at my desk is the best♥ So big thanks to Connectbot for being so good!)

snan:
The issue is that the cursor is white; that's why I can't see it, it's white on white. An older version of ConnectBot supported the cursor being set to a different color.

PR #558, opened 2017-11-08
Fix cursor rendering and visibility

morckx:
After PixelXorXfermode was removed in API 24 and the corresponding upgrade in d372c85dd388176cc2858201f62898b6bad0e42b cursor visibility was rather a matter chance. In addition, the character under the cursor and the metaState symbols were no more visible, in any case.

Emulate the old XOR cursor by using a ColorMatrixColorFilter.

This fixes https://github.com/connectbot/connectbot/issues/533

Signed-off-by: Marc K <morckx@gmail.com>


kruton:
Close/re-open to poke the build bots.

PR #560, opened 2017-11-14
Respect font scale and changed system settings in font size calculation

### LINKED ISSUE #532: Can't select text
Just updated to the new connectbot in F-Droid; I love the _design_ of the new UI for selecting text, but I haven't gotten it to work in practice. I longpress, some letters in the middle of the word are selected. The left handle I can sorta, sometimes, with a couple of tries and restarts move to the correct place. The right handle I can't move at all horizontally. I can drag it down a few lines though, but that doesn't help me (I wanted to copy one very long word).

The old system usually took me a couple of tries and was often frustrating; with the new one I tried for five minutes until giving up and going to my desktop computer.

I love connectbot, it's my most used phone app for sure.

morckx:
* take into to account that the user can also change the font scale
  (under system-settings -> display -> font size)
* take into account that density and font scale can be changed while the app is
  running

This should fix #532 in cases where font scale is not 1.0.

### DIFF
--- a/app/src/main/java/org/connectbot/service/TerminalBridge.java
+++ b/app/src/main/java/org/connectbot/service/TerminalBridge.java
@@ -41,6 +41,7 @@ import android.graphics.Color;
 import android.graphics.Paint;
 import android.graphics.Paint.FontMetrics;
 import android.graphics.Typeface;
+import android.provider.Settings;
 import android.text.ClipboardManager;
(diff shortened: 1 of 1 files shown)


morckx:
Strange, my identical travis [build](https://travis-ci.org/morckx/connectbot/builds/302074900) passed.

PR #561, opened 2017-11-17
Accept colon as SGR sub-parameter separator

whydoubt:
Based on discussion about #531, ':' should work for denoting sub-groups
of SGR parameters.  Add this capability, and check these parameters a
bit more strictly.

### DIFF
--- a/app/src/main/java/de/mud/terminal/vt320.java
+++ b/app/src/main/java/de/mud/terminal/vt320.java
@@ -2526,6 +2526,9 @@ public void setScreenSize(int c, int r, boolean broadcast) {
             DCEvars[DCEvar] = DCEvars[DCEvar] * 10 + (c) - 48;
             term_state = TSTATE_CSI;
             break;
+          case ':':
+            DCEvars[DCEvar] |= (1<<31);
+            /* intentionally fall-through here */
           case ';':
             DCEvar++;
             DCEvars[DCEvar] = 0;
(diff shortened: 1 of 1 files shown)


whydoubt:
If some other people could beat on this to make sure this doesn't break existing use cases, that would be great.

leonerd:
Looks generally about what I had in mind.

A few specifics though:

Some of the variable names might want adjusting. There's a certain conflation of what is being called a "parameter" here, because you're naming the subparams "colon-separated params". If you name them "subparameters" it might make some of the corresponding counter variables a bit clearer.

If you pull out the (first subparam of the) SGR parameter we're currently working on as something like

    int sgr = DCEvars[i];

early enough in the processing part of the code, then that leaves you free to continue incrementing `i` over the subparams, and also lets you use that more directly in later code.

whydoubt:
>There's a certain conflation of what is being called a "parameter" here, because you're naming the subparams "colon-separated params".

I don't believe I did.  However, I will work on improving my terminology.  I am now taking a closer look at ITU-T Rec T.416 (in lieu of ISO 8613-6) and ECMA-48 (in lieu of ISO 6429), even though those documents are not consistent.  I am about to brain-dump some of what I've gathered.

So let's say we have `\x1b[7;1;48:5:20;38:5:60m`...

According to ECMA-48:
- `7;1;48:5:20;38:5:60` is the _parameter string_
- _parameter_ _sub-strings_ are delimited by semi-colons
- so we have four parameter sub-strings: `7`, `1`, `48:5:20`, `38:5:60`
- a colon may be used as a separator within a parameter sub-string
- regarding the details for _parameter_ _values_ `38`/`48`, the document refers you to T.416 / ISO 8613-6

According to T.416:
- we have four _selective_ _parameters_, each with a single _parameter_ _value_
- the parameter values: `7`, `1`, `48`, `38`
- the parameter value `48` is _followed_ by a _parameter_ _substring_: `5:20`
- the parameter value `38` is _followed_ by a _parameter_ _substring_: `5:60`
- the parameter substring is divided by colons into _parameter_ _elements_, and this is the first mention in the document of a separator character relating to parameters
- `5` and `20` are the first and second parameter elements for the parameter value `48`
- `5` and `60` are the first and second parameter elements for the parameter value `38`
- Also: "If the first parameter element has the value 2, 3, or 4, the second parameter element specifies a colour space identifier referring to a colour space definition in the document profile." ... "If the first parameter element has the value 2, the parameter elements 3, 4, and 5, are three integers for red, green, and blue colour components." In other words, according to spec, it should be 48:2:\<colour space id>:\<R>:\<G>:\<B>.  But does _anyone_ treat it that way?
> If you pull out the (first subparam of the) SGR parameter we're currently working on as something like ...snip... early enough in the processing part of the code, then that leaves you free to continue incrementing i over the subparams, and also lets you use that more directly in later code.

I am hazy on what you have in mind.  The first parameter element of the substring (such as `2` or `5`) will determine the position of subsequent elements of the substring.  How does looping over them improve the code?  We would need some additional state variables I think.

I do agree with making (1<<31) a named flag.  It is used several different locations and is somewhat magic.

Issue #458, opened 2016-12-22
Changing fonts


xelra:
It would be great if one could change the font of connectbot. 

Personally I use the font [Hack](https://github.com/chrissimpkins/Hack) because it's a great monospaced font. What makes this a necessity though is that I'm a heavy user of Vim/Neovim and that plugins like air-/powerline demand a few special glyphs to be patched into the font. [Hack](https://github.com/chrissimpkins/Hack) for example already has those glyphs patched which are essential for proper rendering of some advanced features.

DAlexIV:
Implemented here https://github.com/connectbot/connectbot/pull/565

fat-tire:
See an update to #565 - I've implemented downloadable fonts and just need to clean it up and test it and stuff.

<img src="https://user-images.githubusercontent.com/528174/40447943-f354ace4-5e88-11e8-957e-2616dd6045ff.png" width="60%">

<img src="https://user-images.githubusercontent.com/528174/40448215-bee19b6a-5e89-11e8-988a-d63112ba775d.png" width="50%"><img src="https://user-images.githubusercontent.com/528174/40448755-5243e1b4-5e8b-11e8-9d92-730d28ea4cb1.png" width="50%">

raniesantos:
I just installed Connectbot from the Play Store and I can't find this feature.

Has this really been hanging for the last 10 months.

luhux:
I install the connect bot from the Fdroid ,
I can't find this feature.

fat-tire:
Since it's not explicitly mentioned, I should probably note the actual font code is #613.

clmbmb:
@fat-tire I know you're not payed for support, but has this ever been implemented? I've installed ConnectBot from f-droid and play store, but I have no option to choose the font.

fat-tire:
Heh.  this bug was here when Obama was president... and I don't think #613 was ever accepted.  I haven't touched this for years no or even tried connectbot for a while, but if i knew it *would* be accepted I'd be willing to bring it up to date if needed.

GiulioTani:
Is this dead dead? For some reason the font displayed to me is not even monospace and the cursor does not align with the text leaving me to guess where am I going to edit...

PR #565, opened 2017-12-16
Changing fonts

DAlexIV:
Feature was implemented according to https://github.com/connectbot/connectbot/issues/458 request. For now, I've added only Hack font, but expanding font base shouldn't be a big deal, just put .ttf file in assets/fonts folder and add its name in FontManager. 
Also there could be [Downloadable fonts](https://developer.android.com/guide/topics/ui/look-and-feel/downloadable-fonts.html), but we need to upgrade minsdk to 14 to use them in support library.
And last but not least, I've added a font picker in host editing activity, so we need to translate two strings to all supported languages.
Looking for your feedback on this.


app/src/main/java/org/connectbot/util/FontManager.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Locale;
+import java.util.Set;
+
+import android.content.Context;
+import android.graphics.Typeface;
+
+public class FontManager {
+	private static String FONT_DIRECTORY = "fonts";
+	public static String SYSTEM_FONT = "System";
+	private static Set<String> FONTS = new HashSet<>(Arrays.asList( "Hack"));
```
kruton: nit: space before `"Hack"`.

fat-tire:
Just wanted to update this issue-- I'm working on something fun-- now that we're targeting sdk 27-- we can finally do [downloadable fonts](https://developer.android.com/guide/topics/ui/look-and-feel/downloadable-fonts) :)

Screenshots are on Chrome OS (Samsung Chromebook Pro).





(won't post all the mono fonts.  You get the idea.  And don't ask me why I started typing the numbers with "Z" in nano.  That's a typo, not an issue with the fonts.)

Some fonts such as VT323 have kerning turned on.  Is there an easy way to tell a Paint object not to do this?

This is pretty much done, though I haven't really done much testing to make sure everything works as expected.  I did update a bunch of libraries and such at the same time-- so I want to separate those changes out from the minimum required to make this work and verify that it's backwards compatible.  I think it should fallback for the -oss build as well-- the menu item simply won't appear and the default will be monospace.  Can easily incorporated the changes in this PR though if you want to bundle a few, but the downloadable fonts option adds virtually no size to the app (all fonts are cached by google play services).

Cheers.
ft

PS-- @DAlexIV I tried to use a few bits of your PR, but I didn't need stuff like the FontManager since that is handled by the FontRequest, and I had to change some stuff to make it asynchronous anyway to handle situations where for some reason the font isn't available (in which case it falls back to monospace).  I also didn't style the font options because it would require pre-downloading every one of them first.  But if you wanted to add bundled fonts, it would be very easy to do.

tigerinus:
Any update?

fat-tire:
See #613

ebardie:
I use nerd fonts in my shell prompts, tmux furniture, etc., but ConnectBot's single font doesn't place nicely.

I know that switching to `termux`  is an option, but I'd prefer to stick with ConnectBot if only it could use custom fonts.
  
Given that this ticket was closed after a CLA issue, and that maybe the following would be worth considering to help unblock contributions?

https://ossbase.org/initiatives/cla-free/

PR #575, opened 2018-01-18
Update SDK, API, tools, plugins, libraries, etc. to latest in AS 3.1 (and support multiwindow on CB)

fat-tire:
So last night I was using ConnectBot on my Chromebook and got a little annoyed that I couldn't resize it, so figured I'd add that to the manifest and build it-- which lead to me updating the build to the latest stuff in Android Studio 3.1.  To do so I had to add a pretty generic NotificationChannel (feel free to change/move it somewhere else if you want) and some other minor tweaks.  Also the minimum supported API in google-services is now 14, so that's true of the app too.

Don't know if this breaks any tests that weren't breaking before, but anyhoo, I have resizable windows on my Chromebook now so I'm happy.  YMMV.  Lots of tracked .idea files (and other files) changed automatically during the build, but as there's no reason I can tell to track this in the repo, I also added .idea to .gitignore.  Android Studio/lint also recommended lots of other changes (formatting, code simplification, etc.) but I only include the bare minimum needed for the build to complete and hopefully didn't leave anything out.  Obviously someone will have to verify it builds for them too :)

I'm guessing that build daemon thing will break as there are new license things for 27.0.x that need to be agreed to.

This should also clear a path for downloadable font integration re: #565.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ build/
 
 # Android gradle native build
 \.externalNativeBuild/
+.idea/
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -172,9 +172,9 @@ dependencies {
     ossCompile 'org.conscrypt:conscrypt-android:1.0.0.RC14'
 
     testCompile 'junit:junit:4.12'
-    testCompile 'org.mockito:mockito-core:1.10.19'
-    testCompile 'org.assertj:assertj-core:1.7.0'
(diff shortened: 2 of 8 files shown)


fat-tire:
Figured I could update to Java 8 as well... build seems to work,so just added that too.

kruton:
This looks good to me, but it appears to have broken the testing with `./gradlew check` and `./gradlew connectedCheck` which would prevent testing in Travis-CI. That would need to be fixed before merging.

Issue #570, opened 2018-01-02
Disconnect all doesn't work


Choukajohn:
Hi,
I have a bug in last release 1.9.2 on Android 7.1.1 with the function 'Disconnect all'.

How to reproduce the bug:
Launch app, connect to 2 or more hosts, try to disconnect from all hosts with the 'Disconnect all' option in home menu.

To note:
When I am connected to 2 hosts or more, only if I clear ConnectBot from recent apps in Android and then I come back to the app through Android notification and use 'Disconnect all' option, it works.

Tell me if I can help more with a log or something else.

Thanks.

Offerel:
+1

Offerel:
+1

Concipient:
I'm having this same issue.

netman69:
The same on here, in android 7.0

nindanaoto:
This is already resolved by #590.

PR #590, opened 2018-03-21
Disconnect all / disconnect from notification or connections list not working.

netman69:
I had what I believe is the same issue as  #570, this patch fixes it by posting to a new handler on the UI thread instead of the parent view. I do not know why posting to the TerminalView did not work (as far as I can tell it works, but not when on the UI thread already).


netman69:
Apparently I still have to learn using pull requests, only the first commit is relevant to my explanation here. The other commit might be useful also though, it prevents the prompts from being drawn over the virtual keys thing.

PR #608, opened 2018-05-24
Fix a lot of issues, add Error Prone, etc.

kruton:


kruton:
It holds the code style settings among other things.

fat-tire:
I think anything in .idea/ is regenerated/overwritten as-needed by Android Studio (the gradle build scripts)    Delete it and it will all be recreated.  The iml files may be needed though, but I thought they would be recreated too.  Maybe not (?)

PR #610, opened 2018-05-27
Upgrade ClipManager (from deprecated version)

fat-tire:
Out with the bad text-assuming clipboard.  In with the new content-aware clipboard!

Seems to work.  YMMV.


app/src/main/java/org/connectbot/PubkeyListActivity.java
```diff
@@ -58,7 +60,7 @@
 import android.os.IBinder;
 import android.support.annotation.VisibleForTesting;
 import android.support.v7.widget.LinearLayoutManager;
-import android.text.ClipboardManager;
+import android.support.v7.widget.RecyclerView;
```
kruton: Not sure how this snuck in there, but it's unused.

fat-tire:
Dunno what that recyclerview import was about, but it's gone now :)  cheers

PR #612, opened 2018-05-27
Fix deprecated methods, redundant backslashes, string issues and

fat-tire:
other lint identified issues.  Not too much interesting here, but please check and make sure nothing broke.


app/src/main/java/org/connectbot/transport/SSH.java
```diff
@@ -920,11 +919,10 @@ public void setUseAuthAgent(String useAuthAgent) {
 				} else if (privKey instanceof EdDSAPrivateKey) {
 					EdDSAPublicKey pubkey = (EdDSAPublicKey) pair.getPublic();
 					pubKeys.put(entry.getKey(), Ed25519Verify.encodeSSHEd25519PublicKey(pubkey));
-				} else
-					continue;
+				} else {
+                }
```
kruton: This should be tabs.

PR #614, opened 2018-05-30
Fix German F11 label in emulated keys

### LINKED ISSUE #607: Terminal F11 is missing (F10 twice)
Hello,
The F11 Button is missing and the F10 Button is there twice.
Image of F11 is misssing

morckx:
Fixes #607


fat-tire:
Ah, so it was just in the german keys... cool

Issue #599, opened 2018-05-13
Network disconnect on localhost connection


sleirsgoevy:
When ConnectBot detects a network change, it will break the connection and print the "Connection lost" error. It even does that when connected to localhost (127.0.0.1). I think the network state logic should be disabled for localhost connections.

PR #624, opened 2018-07-05
Do not drop connections if Android reports no connectivity

### LINKED ISSUE #623: Disconnects while connected through VPN
I am using Wireguard as a VPN. Wireguard uses UDP and reconnects automatically, TCP connections through it usually stay alive. However, when using ConnectBot connections get disconnected when roaming from WiFi to 4G. This seems to be caused by ConnectBot itself. The ConnectivityManager reports for a short period that there is no connectivity, which causes ConnectBot to kill all connections:

    07-05 01:07:09.438 13773 13773 D CB.ConnectivityManager: onReceived() called; noConnectivity? true; isFailover? false
    07-05 01:07:09.483 13773 13773 D CB.ConnectivityManager: onReceived() called; noConnectivity? false; isFailover? false

Android nowadays uses SOCK_DESTROY to close sockets depending on which interface the TCP has been useing. It seems that manual handling as it is used in ConnectBot is not ideal and no longer necessary.

agners:
Android (netd) kills TCP connections if addresses go away since
Android 5.x (Lollipop). Actively killing connection is unecessary
and in some situations even harmful, e.g. when using localhost
connections or if a VPN is present which will reconnect on its
own.

Fixes #623 and #599.

### DIFF
--- a/app/src/main/java/org/connectbot/service/ConnectivityReceiver.java
+++ b/app/src/main/java/org/connectbot/service/ConnectivityReceiver.java
@@ -79,22 +79,14 @@ public class ConnectivityReceiver extends BroadcastReceiver {
 			return;
 		}
 
-		boolean noConnectivity = intent.getBooleanExtra(ConnectivityManager.EXTRA_NO_CONNECTIVITY, false);
-		boolean isFailover = intent.getBooleanExtra(ConnectivityManager.EXTRA_IS_FAILOVER, false);
+		NetworkInfo info = (NetworkInfo) intent.getExtras()
+				.get(ConnectivityManager.EXTRA_NETWORK_INFO);
 
-		Log.d(TAG, "onReceived() called; noConnectivity? " + noConnectivity + "; isFailover? " + isFailover);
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/connectbot/service/ConnectivityReceiver.java
```diff
@@ -79,22 +79,14 @@ public void onReceive(Context context, Intent intent) {
 			return;
 		}
 
-		boolean noConnectivity = intent.getBooleanExtra(ConnectivityManager.EXTRA_NO_CONNECTIVITY, false);
-		boolean isFailover = intent.getBooleanExtra(ConnectivityManager.EXTRA_IS_FAILOVER, false);
+		NetworkInfo info = (NetworkInfo) intent.getExtras()
+				.get(ConnectivityManager.EXTRA_NETWORK_INFO);
```
kruton: Can you update this to use [`NetworkManager.getActiveNetworkInfo()`](https://developer.android.com/reference/android/net/ConnectivityManager.html#getActiveNetworkInfo()) since this API is deprecated?

PR #680, opened 2018-12-12
Switch to adaptive icons. Fixes #545

### LINKED ISSUE #545: Add support for Android Oreo's adaptive icons
On Android Oreo, a new feature exists called "adaptive icons." This allows the user to select which shape they want all of their icons to appear. For example, on a Pixel device, icons are circular by default, but options exist for "square," "rounded square," "squircle," and "teardrop." 

I believe the easiest way to accomplish this is simply to change the targeted OS to Android Oreo upon compilation. From what I gather, this will automatically put the icon against a white background on supported devices.

[Here's some documentation if you'd prefer to use more customized icons than the default white background.](https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive.html)

sal0max:
I tried to stay as close to the original icon, yet respect the new guidelines about launcher icons.

Preview:
adaptive icons template


kruton:
That looks great. Are the `icon_round` resources needed? It looks like it's not being used from the manifest.

sal0max:
@kruton 
I forgot to add their usage to the `Manifest.xml`, [like one should do](https://developer.android.com/about/versions/nougat/android-7.1#circular-icons).
Could add this in a further pull request.

kruton:
Ah, it seems there are [conflicting pieces of advice](https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive#creating_adaptive_icons_in_xml):

> You can also define an icon drawable resource using the android:roundIcon attribute. You must only use the android:roundIcon attribute if you require a different icon asset for circular masks, if for example the branding of your logo relies on a circular shape.

sal0max:
There seems to be a lot of confusion with the three types of icons: regular, round and adaptive.

Ian Lake of Google [writes](https://medium.com/@ianhlake/on-a-device-that-uses-roundicon-that-roundicon-will-always-be-used-youll-note-that-there-are-ef59a2df4c60), that `roundIcon` is something a devices build configuration would choose. That's the same as written in the 7.1 changelog I linked.

As I see it, this means if `iconRound` is present and set in the `Manifest.xml`:
* Devices < API25 will always choose the regular png version.
* Devices = API25 (Android 7.1) choose either the regular or the round png version, depending on the build configuration.
* Devices ≥ API26 (Android 8.0) use either the adaptive version or the round adaptive version. 
   Ian includes a round adaptive icon, that is only a reference to the regular adaptive one:
   ```xml
   <resources>
        <mipmap name="ic_launcher_round">@mipmap/ic_launcher</mipmap>
   </resources>
   ```
   That's something I'd also change when adding `iconRound` to the manifest.

I personally don't really care if `iconRound` is present. My devices don't make use of that feature. In fact, I doubt that any high number of devices make use of that feature.
I just added it to the pull request, as Android Studios Image Asset Studio produces round icons by default.

So, I'd make it up to you whether round icons should be included or not. :)

PR #707, opened 2019-07-29
added a fdroid-badge

DJCrashdummy:
...according to https://gitlab.com/fdroid/artwork/blob/master/badge/README.md


BoFFire:
Hi, I think it is a good idea to add the F-Droid badge. Thank you!

DJCrashdummy:
@kruton why not finally merge this (or something similar)?

Issue #710, opened 2019-08-19
NPE on org.connectbot.service.TerminalBridge#onDraw


lhchavez:
## Bug description

[`org.connectbot.service.TerminalBridge#onDraw`](https://github.com/connectbot/connectbot/blob/c4f21c10af8b0abde12ccfc3f20060362fe9e124/app/src/main/java/org/connectbot/service/TerminalBridge.java#L714) can sometimes crash with an NPE while trying to access character attributes at https://github.com/connectbot/connectbot/blob/c4f21c10af8b0abde12ccfc3f20060362fe9e124/app/src/main/java/org/connectbot/service/TerminalBridge.java#L733

This is caused by [`de.mud.terminal.VDUBuffer#insertLine(int, int, boolean)`](https://github.com/connectbot/connectbot/blob/c4f21c10af8b0abde12ccfc3f20060362fe9e124/app/src/main/java/de/mud/terminal/VDUBuffer.java#L335) not always filling up all character/attribute rows.

## Steps to reproduce

1. Connect to a Ubuntu Bionic 18.04.2 LTS host
2. Open tmux
3. Start fiddling with zoom and scroll until it crashes

## Expected behavior

It shouldn't crash

## Android device

 - Device: Android emulator (google_apis_playstore/x86)
 - OS: Android 9 PSR1.180720.075
 - ConnectBot Version: 1.9.5, as well as master.

## Server information

 - OS: Ubuntu Bionic 18.04.2 LTS
 - SSH Software and Version: OpenSSH 7.6p1
 - Tmux version: 2.6

## Additional context

One sample crash trace:

```
08-17 18:23:40.270 10152  8683  8683 E AndroidRuntime: FATAL EXCEPTION: main
08-17 18:23:40.270 10152  8683  8683 E AndroidRuntime: Process: org.connectbot, PID: 8683
08-17 18:23:40.270 10152  8683  8683 E AndroidRuntime: java.lang.NullPointerException: Attempt to read from null array
08-17 18:23:40.270 10152  8683  8683 E AndroidRuntime: 	at org.connectbot.service.TerminalBridge.onDraw(TerminalBridge.java:733)
08-17 18:23:40.270 10152  8683  8683 E AndroidRuntime: 	at org.connectbot.TerminalView.onDraw(TerminalView.java:423)
08-17 18:23:40.270 10152  8683  8683 E AndroidRuntime: 	at android.view.View.draw(View.java:20207)
[...]

PR #711, opened 2019-08-19
Ensure that the VDU buffer's lines are never null

lhchavez:
This change makes it such that the VDU buffer's lines are always defined
(although they might be empty).

Fixes: #710

### DIFF
--- a/app/src/main/java/de/mud/terminal/VDUBuffer.java
+++ b/app/src/main/java/de/mud/terminal/VDUBuffer.java
@@ -359,8 +359,8 @@ public class VDUBuffer {
       if (n > (bottom - top)) n = (bottom - top);
       int size = bottom - l - (n - 1);
       if(size < 0) size = 0;
-      cbuf = new char[size][];
-      abuf = new long[size][];
+      cbuf = new char[size][width];
+      abuf = new long[size][width];
 
       System.arraycopy(charArray, oldBase + l, cbuf, 0, bottom - l - (n - 1));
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/mud/terminal/VDUBuffer.java
```diff
@@ -359,8 +359,8 @@ public synchronized void insertLine(int l, int n, boolean scrollDown) {
       if (n > (bottom - top)) n = (bottom - top);
       int size = bottom - l - (n - 1);
       if(size < 0) size = 0;
-      cbuf = new char[size][];
-      abuf = new long[size][];
+      cbuf = new char[size][width];
```
lhchavez: i was debating between doing this and L391 (for `cbuf`) vs. adding something like

```java
for (int i = 0; i < cbuf.length; i++) {
    if (cbuf[i] != null) {
        continue;
    }
    cbuf[i] = new char[width];
    Arrays.fill(cbuf[i], ' ');
}
```

just before [L477](https://github.com/connectbot/connectbot/blob/c4f21c10af8b0abde12ccfc3f20060362fe9e124/app/src/main/java/de/mud/terminal/VDUBuffer.java#L477). `abuf` is fine since it will be zero-initialized. Let me know what you prefer.

lhchavez:
Also, I'm waiting on go/ospo's approval before signing the CLA.

ilf:
It would be super awesome to get this fixed and merged (and released). Thanks!

Issue #702, opened 2019-06-19
Incorrect terminfo data?


geometryof:
Just installed ConnectBot 1.9.5 under Android 9 -- after many years of using 1.7.1 under 4.4.
Now, when I SSH to a Linux host -- multiple servers running different flavors -- I get a TERM of xterm-256color, rather than the old screen.  Works fine with bash, but not with tcsh.  The cursor placement is now off, e.g. moving to the beginning of the line (ctrl-a) appears to put the cursor over the second character, as does clearing the line (ctrl-u); in both cases, a refresh (ctrl-r) fixes the display. Changing the TERM to screen fixes the problem, but of course at the cost of other xterm functionality.  That's probably OK for me, but it looks like a terminfo bug somewhere.

I would welcome suggestions (other TERM alternatives to try?) or comments.

Thank you.  (And thank you for ConnectBot, an essential tool.)

nindanaoto:
This seems to be fixed in the main. If we change the term type in the profile, `$TERM` is changed.

PR #755, opened 2020-03-17
Fix an off-by-one error in ESC [ G handling

tirsek:
There's an off-by-one error in the terminal emulator's handling of the ESC [ G (Cursor Horizontal Absolute) sequence.

I believe this also solves issue #702.


tirsek:
Oops, committed using the wrong e-mail. Let me try that again... :-/

PR #850, opened 2020-09-04
Update to sshlib 2.2.15

kruton:


rhardy613:
Thank you for fixing this. These fixes are critical for many users to be able to use the client. 
It's been almost a year since the last update. Please consider releasing a client update which identifies it contains the fixes through Google Play. If there is something preventing use of Google Play please an alternative such as F-Droid or as last resort a side load build would make use possible again.

kruton:
There is an update pending, but it causes errors using Ed25519 on some
devices which appear to have a broken ART implementation. I was working
with the implementors of the crypto library to figure out what is happening.

rhardy613:
Please excuse the process ignorance. Is there a way try that pending update to see if it works on my One Plus 7T?
So far all I've found that git-v1.9.7-36 or later on the github releases download page should have the fix but there seems to be no description any where on what the difference is between the OSS and Google versions of the APK is. 
Is one signed for Google Play and the other for F-Droid or some other market?

kruton:
One is for devices that have Google Play Services ("Google") and the other
("oss") doesn't need Google Play Services. There is a slight size penalty
for not being able to use Google Play Services, but they are otherwise
equivalent.

eworm-de:
Wondering if this solves my issues connecting to Mikrotik devices...
Would you be so kind and check to connect to `admin@demo.mt.lv` (without password)? Thanks a lot!

PR #938, opened 2021-05-17
Keep codestyle only

hannesa2:
It keeps default codestyle from Android Studio 4.2

There was too much pointless stuff checked-in which prevents me to build on Apple silicon with right java


kruton (approved):

kruton:
This is going to be painful for all the folks that have this checked out already, but that's probably just me.

PR #1105, opened 2022-04-01
provide method to delete known host entries for ssh hosts

### LINKED ISSUE #477: No way to update known_hosts key after host key change
I generated new private keys on some servers. This is a reasonable measure to take when it's possible that keys have been compromised.

Now, every time I connect from connectbot, I get a warning that the key has changed. I would like the warning to go away, and for the known_hosts to be updated such that if someone REALLY DID man-in-the-middle me, I would actually know about it, going forward!

### LINKED ISSUE #586: Can't delete old known_hosts
My host changed, and now ConnectBot won't let me connect.  The connection is immediately dropped -- presumably since the saved key in ConnectBot doesn't match the server end.
There is no way to delete the saved key, or overwrite it with the new one.

richteas75:
to be used if the server key or algorithm was changed
A new menu entry is added in the HostList - when executed,
all known keys in the database for the corresponding host are
deleted.
resolves #477
resolves #586

### DIFF
--- a/app/src/main/java/org/connectbot/HostListActivity.java
+++ b/app/src/main/java/org/connectbot/HostListActivity.java
@@ -499,6 +499,36 @@ public class HostListActivity extends AppCompatListActivity implements OnHostSta
 					return true;
 				}
 			});
+			MenuItem deleteFromKnownHosts =menu.add(R.string.list_host_reset_knownhost);
+			deleteFromKnownHosts.setOnMenuItemClickListener(new OnMenuItemClickListener() {
+				@Override
+				public boolean onMenuItemClick(MenuItem item) {
+					// prompt user to make sure they really want this
+					new androidx.appcompat.app.AlertDialog.Builder(
+							HostListActivity.this, R.style.AlertDialogTheme)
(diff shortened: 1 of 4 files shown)


mathias-ch1903:
I ran into the same issues, it seems that the feature is still missing in the latest 1.9.9 version, but I was not in the mood to recreate all my host entries.

The following procedure was working for me at least (root needed)
- Stop the app connectbot
- Copy the file /data/data/org.connectbot/databases/hosts to a user location
- Open the sqlite db "hosts"  with an sqlite editor (ie. aSQLiteManager, old but still working)
- remove all entries from the table "knownhosts"
- copy the "hosts" file back to the original location, overwrite
- start connectbot again, connection is possible.

kruton:
This needs a test or else the functionality will be accidentally removed or broken in the future.

kruton (changes requested):
This needs tests or else the functionality may be broken or removed inadvertently in future changes.

nindanaoto:
I believe we can close this PR now, as #1814 implements the proposed functionality.

PR #1112, opened 2022-04-18
Simplify NDK

hannesa2:
With it you can build it fast on an Apple Silicon M1 too

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -43,7 +43,6 @@ android {
 
 		minSdkVersion 14
 		targetSdkVersion 29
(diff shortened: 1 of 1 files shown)


kruton:
Does this remove any ABIs from the build?

kruton (approved):

hannesa2:
No it keeps it like it is. During debug only target architecture is included.
Does the currently really filter something ? There are all included, except historical `mips`.

On a play store release I recommend to use aab instead of apk publish. This filters all libraries to used hardware

PR #1174, opened 2022-10-02
MinSdk 21 to solve multidex

hannesa2:
I run into multidex issues. With minSdk 21 you reach 98% of all Android devices. 
Devices < 21 can still use recent version.


kruton:
There's something up with Android Gradle Plugin 7.2 that causes the dex file to go from 2.3MB to 6.2MB. This requires further investigation instead of just turning on multidex and taking the size hit.

kruton:
Okay, this appears to be fixed with #1222

PR #1205, opened 2023-01-22
Fix text selection for certain font sizes

proski:
For certain font sizes (e.g. for 13 points), the overlay was using slightly smaller font than the terminal bridge. It appears that the overlay (i.e. TextView) is always rounding down the pixel size, whereas TerminalBridge is rounding to the closest integer.

Copy the pixel size from the bridge to the overlay to ensure perfect alignment between the two.

Remove TerminalBridge#setFontSize(), it's not used anymore. Add TerminalBridge#setFontSizePx() instead.

### DIFF
--- a/app/src/main/java/org/connectbot/TerminalView.java
+++ b/app/src/main/java/org/connectbot/TerminalView.java
@@ -48,6 +48,7 @@ import android.net.Uri;
 import android.os.AsyncTask;
 import android.preference.PreferenceManager;
 import android.text.ClipboardManager;
+import android.util.TypedValue;
 import android.view.GestureDetector;
 import android.view.KeyEvent;
 import android.view.MotionEvent;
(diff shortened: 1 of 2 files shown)


proski:
One way to check the alignment is to make the overlay text visible by replacing `setTextColor(Color.TRANSPARENT);` with `setTextColor(Color.RED);` in `TerminalTextViewOverlay.java`. Use volume keys to change the font size and touch the text to see the alignment.

proski:
I believe the code can be greatly simplified and made more robust by changing the fonts size for the overlay first and then copying the `Paint` settings from the overlay to the bridge. All font size conversions would be made by Android. But that would be a much bigger change. The overlay would determine the screen layout, the bridge would adapt to it. The font size in points won't need to be in the bridge code, it will be exchanged between the overlay and the host bean.

kruton (approved):

kruton:
Thanks for the fix. The more extensive fix would be welcome as well.

PR #1219, opened 2023-02-08
Update AGP and Gradle

kruton:


kruton:
This can't be updated because of usefulness/easylauncher-gradle-plugin#408

PR #1342, opened 2023-12-23
Enable pinch to change font size

### LINKED ISSUE #430: Add pinch-to-zoom

morckx:
Standard way of changing the font size on touch screen devices. It can be easily and flexibly handled (e.g. with changing lighting conditions) and doesn't interfere with volume settings nor with selecting text. Two finger touches are reliably distinguished from one finger long presses.

Resolves #430


morckx:
I'm running this successfully since my initial version of this PR in 2016 :smile:

kruton:
One of the changes that non-Latin alphabet users have asked to see is
resizing text with pinch zoom but not changing the window size (e.g., text
gets bigger but it stays at 80x25).

morckx:
> One of the changes that non-Latin alphabet users have asked to see is resizing text with pinch zoom but not changing the window size (e.g., text gets bigger but it stays at 80x25).

You mean so that the actual device gets bigger and smaller when you pinch? Yes, I'd love that too.

kruton:
I've created https://github.com/connectbot/termlib which allows you to pinch to zoom sections. It's currently used in the beta version, but will be in the v1.10 and later releases.

Issue #1128, opened 2022-05-14
Post-login automation does not send my command, merely types it in, for Windows


Firsh:
The server is Windows OpenSSH. Anything, even just
`dir`
put into post-login automation just gets copy pasted into the terminal but Enter is not pressed for me. It works on Ubuntu servers. Is there a special way of ensuring the command gets sent, or is this a bug? I tried newlines and ; character.

skingzd:
me too.Use it connect to windows,always need to press enter manually .

nolddor:
Same here connecting via SSH to Debian server.

@kruton any chance to look into this?

ckujau:
Duplicate of #1097? Although I've experienced the same for years now, not sure why both these issues were only opened recently.

atikattar1104:
Same For Termux As Well

lordofthemoon:
I've just started using ConnectBot to do this, and I see the same thing.  I'm connecting to a Raspberry Pi running raspbian stretch, using a keypair so that I don't have to use a password.  I just want to execute a single command on the remote box and then disconnect, but the command in "post-login automation" just gets placed on the command line, and I have to press enter to actually execute it. (v1.9.9 according to menu > Help).

Alteus:
Two years later, and this issue is still open. As the matter of fact, the solution is trivial, just hit enter key after your command in the post-login automation box.

quentinkrammer:
I had the same issue on a debian server. I solved it by appending a newline character to the end of the shell command.

Example:
All I want is for my Debian server to shutdown as soon as I connect via SSH.

This Post Login-Automation script would require me to manually press Enter after connecting: 
`sudo poweroff`

Whereas this script gets immediately executed: 
```
sudo poweroff

```

Firsh:
Yes, new line (an empty line at the end in the textbox) works fine on Windows SSH as well as Ubuntu now.

atikattar1104:
Maybe This Issue Can Be Closed Now.

Roberto-Gentili:
> Yes, new line (an empty line at the end in the textbox) works fine on Windows SSH as well as Ubuntu now.

Not working for me on Windows SSH

atikattar1104:
Try Adding 2 New Lines. Sometimes, Only One New Line Can Be Considered As False Positive By The Systems.

Roberto-Gentili:
I had already tried with more lines but it didn't work

PR #1384, opened 2024-03-30
Send post login string as SSH command when no shell session is requested.

s3lph:
There is currently no way for ConnectBot to send a command to a SSH server without allocating a PTY. The "post login string" is only ever sent if a PTY is requested and allocated, and then it is simply sent as an input to the PTY. Issue #1128 describes a similar, if not identical issue.

This is especially limiting if a remote SSH server only permits executing commands. Probably the most common example is OpenSSHd's `ForceCommand` option: It requires any command (or a shell) to be started before ForceCommand takes effect. ConnectBot, however, won't start a shell unless PTY allocation succeeds.

This PR adds the following feature: If establishing a session is disabled, but a post login string is provided, the post login string is executed as a command through SSH, rather than being piped into a shell. The local and telnet transports' behavior is unchanged.

### DIFF
--- a/app/src/main/java/org/connectbot/service/TerminalBridge.java
+++ b/app/src/main/java/org/connectbot/service/TerminalBridge.java
@@ -393,6 +393,27 @@ public class TerminalBridge implements VDUDisplay {
 	}
 
 	/**
+	 * Send a noninteractive command to the transport. Used for post-login strings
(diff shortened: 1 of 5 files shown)


kruton:
This doesn't currently compile. It also needs a test to make sure that the functionality isn't broken or removed in future refactors.

PR #1433, opened 2024-07-15
Feature: long press to copy URL from the scanning dialog box

shitlime:
Long click to copy URL from the scanning dialog box.


kruton:
Should have a test or else the functionality might accidentally get deleted during refactoring.

PR #1450, opened 2024-11-11
Fix themed app icon

nhoeher:
Previously, the monochrome version of the icon wasn't referenced correctly from within `icon_round.xml` leading to the themed icon not working correctly in some launchers.


kruton (approved):

Richard5A (approved):
Looks good

PR #1460, opened 2024-12-10
Mosh Implementation

bqv:
I have absolutely no expectation this will get merged. For anyone that wants it however, and for my own future reference, here's a patch I recreated after I lost it two years ago. 

### Embedded Mosh on api29+


kruton:
Oh, this is interesting actually. I wonder if there is a way to split this out so there is no requirement to have a certain ABI.

bqv:
AFAIK those are all the ABIs possible to use on android anyway. It does have to be compiled per-architecture, unless someone is enthusiastic about reimplementing mosh on the jvm... ([llvm](https://github.com/mridoni/llxvm) [magic](https://github.com/davidar/lljvm)...?)

kruton:
Right now you can do a split per ABI, but I don't know if it allows you to just not have the native libraries. The ideal situation is that Android supports wasm! 🥴

kruton:
I'll merge this in for the v1.10 release. The APK size is going to increase anyway, but we can start uploading split ABIs to help with downloads on non-Google Play distribution platforms.

kruton:
It's taking a little longer than anticipated. I'll have to move it to a later release, so I'll put it in v2 for now.

kruton:
Thanks, @bqv , #2169 is merged.

Issue #1098, opened 2022-03-16
F1-F10 keys are not working on Samung DeX


danergo:
On Samsung DeX, Connectbot is an awesome SSH client.

However, I have noticed that in case of using hardware keyboard with Samsung DeX, neither function key is working (F1 to F10).

This would be a great function, as we are all used to use these keys in some command line tools (like mc).

danergo:
ping

danieldjewell:
I'm curious -- what software keyboard do you have selected when this happens? DeX is a little weird because the on-screen/touch keyboard can influence the behavior of physical keyboards in DeX mode.

danergo:
When I use DeX I hook it up to a display and a USB keyboard. No software keyboard is involved, but if you mean in general which software keyboard is _enabled_ and set as _default_ it's only Samsung stock keyboard, and nothing else.

VA1DER:
I have yet to find a soft or hard keyboard where function keys work directly in Connectbot.  Function keys only seem to work when simulated.  Unfortunately, Connectbot simulates function keys differently with soft and hard keyboard.  With soft keyboards you have the option to simulate with CTRL-#.  With hard keyboards you only have the option to simulate with SHIFT-#.

I have tried three different devices, two different hard keyboards, and the Unexpected soft keyboard (which is one of the few soft keyboards which directly support function keys).

@danergo I recommend you re-title this bug to remove "On Samsung DeX" because I do not believe function keys are working directly with any soft/hard keyboard in Connectbot.

PR #1488, opened 2025-04-14
 #1098 - Handle harware keyboard function key

ArnaudChirat:
Translate hardware keyboard function key into VT320 function key code. 
The translation is done if shiftNumbersKeyAreFunction is false or if controlNumberAreFunction is false


kruton:
This is now handled by libvterm in the new Terminal component.

PR #1489, opened 2025-04-14
SSH password reuse

ArnaudChirat:
Allow an user to saved a password when editing a host.
This will bypass the prompt if the password is valid otherwise ask the password


kruton:
Thanks for the contribution. I merged a similar feature.

PR #1513, opened 2025-05-06
Add Android-side text entry feature

kanzure:
Got typos?

During most of my phone use I have long since stopped caring about typos, but over SSH I find that my typo rate is a lot worse and the worse the typo then the more the typo matters.

If you often make elaborate typos with ConnectBot and the Android virtual keyboard, then you might want to try this "text entry" feature. It's a simple enough idea: where you would normally type into the virtual keyboard tied to an SSH session, instead type into an Android native text entry field element. Here you can use the virtual keyboard as you normally would including any glide typing or dictation or other features. When you're done, you can submit your input all at once through the active SSH session.

An alternative to merging this pull request or using this feature is to instead simply open another Android app and type your text input there. Once you are done, copy your text and then paste it into an active ConnectBot session. For me this tends to be faster than typing directly into a ConnectBot session. It also allows long-form input, speech-to-text dictation, and other modern niceties.

### DIFF
--- a/app/src/main/java/org/connectbot/ConsoleActivity.java
+++ b/app/src/main/java/org/connectbot/ConsoleActivity.java
@@ -138,6 +138,7 @@ public class ConsoleActivity extends AppCompatActivity implements BridgeDisconne
 	private MenuItem portForward;
 	private MenuItem resize;
 	private MenuItem urlscan;
+	private MenuItem textEntry;
 
 	private boolean forcedOrientation;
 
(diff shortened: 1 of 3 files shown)


andrewleech:
Hi, I'm interested in adding a feature just like this! Could you maybe add some screenshots here what it looks like in use and how to trigger it?

nindanaoto:
This seems to be already implemented in the main, if my understanding is correct. I believe we can close this PR.

PR #1553, opened 2025-08-16
Add WiFi hotspot port forwarding with bind address selection

steffen-heil-secforge:
* Add WiFi hotspot port forwarding with bind address selection

This adds support for binding port forwards to specific network interfaces, particularly targeting WiFi hotspot scenarios where users want to forward ports only on the hotspot interface.

Features:
- Three bind address options: localhost, all interfaces (0.0.0.0), WiFi hotspot only
- Automatic WiFi hotspot IP detection using interface name patterns
- Port forwarding information displayed directly in host list entries
- Icon tap for quick connect/disconnect without switching to console view

Technical implementation:
- Database schema upgrade (v25→v26) to add bindaddr column
- New NetworkUtils class for WiFi/cellular interface detection
- Graceful handling of devices that don't support AP broadcasts

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -21,6 +21,7 @@
 	<uses-sdk />
 
 	<uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+	<uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
 	<uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
 	<uses-permission android:name="android.permission.INTERNET" />
 	<uses-permission android:name="android.permission.VIBRATE" />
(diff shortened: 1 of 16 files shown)


kruton:
I don't know about this. It seems like it should use [TetheringManager](https://developer.android.com/reference/android/net/TetheringManager#registerTetheringEventCallback(java.util.concurrent.Executor,%20android.net.TetheringManager.TetheringEventCallback)) but that's a brand new API.

steffen-heil-secforge:
Right, but that is API level 36...

~~For now PLEASE DO NOT MERGE!
I am unsure, if I found a bug today and will investigate over the weekend.
I will report back here.~~

I rechecked and my observation was wrong. The unexpected case was caused by not unlocking the private key first, so it was correct that ports were not forwarded.

steffen-heil-secforge:
I think switching to the TetheringManager should be done in the future, when API level 36 is more common.

However, if you prefer I can test the new API and update the logic to use it.

steffen-heil-secforge:
@kruton  Is there a chance this will be merged? Or shall I switch to TetheringManager?

PR #1557, opened 2025-08-27
Add support for specifying IP version

ppb2020:
Added UI and code to support specifying IP version to be used for connections. Defaults to both IPv4 and IPv6. Depends on a similar change in connectbot/sshlib.


nindanaoto:
This can be closed since #1883 is merged.

PR #1671, opened 2025-12-05
ProxyJump

### LINKED ISSUE #679: Support for jump hosts / ProxyJump
Would it be possible to add support for jump hosts, similar to the [`ProxyJump`](https://man.openbsd.org/ssh_config#ProxyJump) directive in SSH? Currently, the only way to implement this is to create a connection with port forwarding, and use that as the target for a second connection. This becomes extremely unwieldy as the number of real targets grows.

Real life example: I currently administer 4 hosts, which only open SSH to one management host. Meaning I would have to create 4 proxy connections with port forwarding, in addition to the 4 final connections, instead of specifying just the jump host for each.

nindanaoto:
Dear maintainers,
This PR resolves #679, adding the ProxyJump capability to this app. 
As you can see in the commit comments, I've used Claude Code for implementing this functionality. Please also note that my primary expertise is in C++, CUDA, and Chisel, rather than Kotlin/Android. 

I don't know what happens if we change the 1.json directly, so I resorted to defining a new schema, 2.json, for this PR. 
Most changes here involve adding ProxyJump logics (e.g., JumpProxyData.kt and StreamSocket.kt). However, I also modified SSH.kt by removing code from tryPublicKey. This was done to extract the getOrUnlockKey function, which is now shared with authenticateJumpHost to allow unlocking the key for the ProxyJump server.

### DIFF
--- a/app/src/main/java/org/connectbot/data/ConnectBotDatabase.kt
+++ b/app/src/main/java/org/connectbot/data/ConnectBotDatabase.kt
@@ -19,6 +19,7 @@ package org.connectbot.data
 
 import android.content.Context
 import androidx.annotation.VisibleForTesting
+import androidx.room.AutoMigration
 import androidx.room.Database
 import androidx.room.Room
 import androidx.room.RoomDatabase
@@ -50,7 +51,8 @@ import org.connectbot.data.entity.Pubkey
  *
  * Migration Strategy:
  * - Version 1: Initial Room schema (migrated from HostDatabase v27 + PubkeyDatabase v2)
- * - Future versions: Schema changes handled by Room migrations
+ * - Version 2: Added jump_host_id column for ProxyJump support (AutoMigration)
+ * - Future versions: Use Room AutoMigration when possible for simple schema changes
  *
  * Security Considerations:
  * - Pubkeys table supports per-key backup control via allowBackup field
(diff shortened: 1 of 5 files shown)


nindanaoto:
I'm sorry that the UI language is Japanese, but I believe you can see the layout in the configuration menu. Satolab is the ProxyJump host server in this case.

nindanaoto:
I found that nested ProxyJump support is missing in my previous commits, so I've added it.  It is now working correctly.

nindanaoto:
Just rebased to the latest main.

nindanaoto:
As above, I believe that a new commit resolves all reviews, and I currently have no unresolved known issues for this PR.

cliomanremi:
@nindanaoto, thanks a lot !
As i'm not a coder...i don't know how to transform your "PR" into an app that i could install in my smartphone in place of the old connectbot version.
Could you please :
- Either describe how to get the app
- Or, simply attach it that we (all people in same situation than me) could download and install it ?
Thanks again for your great job ^^

nindanaoto:
I'm currently building locally and using ADB to install it. 
However, I've noticed that this repository's GitHub Actions include APKs as artifacts. For this PR case, this might be the corresponding one.  
https://github.com/connectbot/connectbot/actions/runs/20029685769
I guess that you can download and install it, though I'm not sure which version (debug/release builds and OSS/Google versions) is built on GitHub Actions.  
If it does not work, I recommend that you make an issue. I agree that if we can get a nightly build from GitHub Actions, that is better.

cliomanremi:
Thanks !

nindanaoto:
I found that the release includes some nightly builds. It might be more handy than accessing GitHub Actions.
https://github.com/connectbot/connectbot/releases

PR #1673, opened 2025-12-06
Disconnect from menu

### LINKED ISSUE #1657: Disconnect host, from host menu

nindanaoto:
This fixes #1657. As a minor change, I added the LinkOff icon to the menu entry and the existing disconnection entry on the terminal page to make them consistent with the other entries. 
I've tested this with #1671 as my daily SSH client from Android, and it is working. 

As you can see from the commits, I've used Claude Code to write this PR.


kruton:
Thanks!

kruton:
By the way, have you been using the new terminal emulator version? What do you think?

nindanaoto:
The only serious issue I noticed immediately with the new termlib was the malfunctioning Ctrl key, but it was fixed 2 days ago, so it is now working correctly.
​Regarding multibyte character support, it made copying Japanese characters easier from the terminal (I'm Japanese). It is definitely better for East Asian users.
​I'm using my forked version based on the current main branch for daily use to include #1671. If I encounter any issues with the new termlib, I will report them.

PR #1676, opened 2025-12-07
Biometric (fingerprint) key

### LINKED ISSUE #1317: Biometric authentication support for keys
To keep the keys more secure and easy to use, we should add support for fingerprint or face id based authentication mechanisms to use the keys for auth. 

Like termius, I should be able to generate a key with my fingerprint as auth and when I log in to the server, it should ask for authentication using the sensor.

I am not experienced in Android, but I would like to contribute for the cause.

nindanaoto:
Dear maintainers,

This resolves #1317 by using the Android Key Store. This PR enables the use of fingerprint-based, encrypted RSA and ECDSA keys, similar to the functionality implemented in [Termius](https://termius.com/).  
As you may notice, getOrUnlockKey is used as introduced in #1671. This is because my daily use version already includes #1671, and I backported that change to remove ProxyJump support, allowing these to be separate PRs.

### DIFF
--- a/app/src/main/java/org/connectbot/service/TerminalManager.kt
+++ b/app/src/main/java/org/connectbot/service/TerminalManager.kt
@@ -477,6 +477,7 @@ class TerminalManager : Service(), BridgeDisconnectedListener, OnSharedPreferenc
 	 * Add a biometric key from Android Keystore to in-memory cache.
 	 * The PrivateKey from Keystore is a proxy that delegates signing to secure hardware.
 	 * Since biometric auth was just completed, the 30-second signing window is active.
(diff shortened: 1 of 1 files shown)


nindanaoto:
I believe that the new commit resolved all reviews. However, I noticed that this PR still has a issue to be fixed: AndroidKeyStore allows the access to the key for only 30 seconds (if my understanding is correct), but the app treats it as unlocked on memory for unlimited time. I will fix that soon.

nindanaoto:
Resolved the relock timeout behavior and rebased to main. Now ready to merge.

PR #1686, opened 2025-12-08
MINOR FIX: Fix deprecated method warnings across codebase

nindanaoto:
I noticed that this app uses multiple deprecated methods, and Gradle is complaining about them. This PR just fixes them. 

- ConnectivityReceiver: Replace deprecated NetworkInfo/BroadcastReceiver with modern NetworkCallback and NetworkCapabilities APIs
- TerminalKeyListener: Update to use android.content.ClipboardManager instead of deprecated android.text.ClipboardManager
- TerminalManager: Replace deprecated bridges property with bridgesFlow.value
- MigrationScreen: Add file-level suppression for LocalClipboardManager
- TerminalKeyboard: Use AutoMirrored icon variants for arrow keys
- HostListViewModel: Replace deprecated disconnected with disconnectedFlow
- PortForwardEditorDialog: Update menuAnchor() to use new overload with ExposedDropdownMenuAnchorType parameter

### DIFF
--- a/app/src/main/java/org/connectbot/service/ConnectivityReceiver.kt
+++ b/app/src/main/java/org/connectbot/service/ConnectivityReceiver.kt
@@ -30,6 +30,12 @@ import android.util.Log
 /**
  * Broadcast receiver that monitors network connectivity changes and manages WiFi locks.
  *
+ * NOTE: Using deprecated BroadcastReceiver/NetworkInfo APIs is intentional.
+ * Migration to NetworkCallback requires thorough testing to ensure connectivity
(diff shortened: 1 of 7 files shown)


kruton (commented):
In general this should be split into several commits so `git bisect` will work if it breaks something.

nindanaoto:
Since my original intention is just to reduce the warnings as a minor fix, I just reverted some problematic changes and added comments for them. For the suppressed file, I rewrote the fix. If this PR ended up just being a bunch of adding comments or leaving the intention descriptions here, it would be fine.

kruton (approved):

PR #1089, opened 2022-02-23
Import&export of hosts and port forwards using Gson

Craeckie:
Fixes #527 and #922.

The hosts and port forwards are exported to the clipboard in json format. This does not include any keys.

The class `HostExport` is a crude workaround. Currently there is no Object-Oriented data structure that maps port forwards to hosts. If someone is willing to refactor the existing code, this class could be removed. This workaround was the simplest way I found to serialize all hosts with their respective port forwards.

Issue #922, opened 2021-04-16
Duplicate profile


nab-os:
Being able to duplicate an existing profile would be practical.
For example in the popup after a long press event on a profile, simply having a "Duplicate" action that just copies all the properties of the selected profile and open the create profile view with it would be nice.

I have a lot of almost identical servers and this feature would really speed up the configuration phase.

kot7k:
I would also want this feature and/or the option to export all/part of your servers to a config file wich you could import again in same/other device.

Issue #527, opened 2017-05-28
Add ability to export/import config


JohnVeness:
Please can you consider adding a way to export and import Connectbot config, ie settings, colours, host configs, keys. I'm aware these can be backed up via Google servers but I don't trust them, and my device isn't rooted to allow use of Titanium backup or similar.

Use cases include moving to a new device, or wanting to restore config after uninstalling/reinstalling Connectbot, e.g. if moving between versions signed with different keys.

Offerel:
+1

cosmospuglisi:
+1

vmario89:
hey. same question for me. i have tons of ssh connections and i want to backup them on another device to be able to re-import them to another android device. my last phone was defect and i had to re-create those connections manually because my phone backup was not compatible with the new phone. It also makes sens to import/export config to give other admin persons in your group the possibilty to quickly share connections

CarloDePieri:
+1 I, too, think it would be a neat feature to help transition to a new device.

hollysheep:
Is this feature inplemented or is it planed?

evanushechkin:
+1

mywkaa:
+1

derflasher:
+1

nab-os:
+1

Matatu:
Still an issue.

faan11:
+1. I also need this feature for the aforementioned purpose

BrandonGoodman:
+1

matzzzl:
+1

DeprecatedHardware:
This is my preferred SSH client on android and the ability to backup/restore host config would be fabulous. I think it is the only feature missing from this application.  Would the devs release a paid for version with this functionality?

gnanet:
An import/export of the connections in some externally editable format could allow us to better manage them, in case some general parameter changes for a larger group of connections, it would not be a PIA to edit them.

skeezmoe:
Please please add an Export/Import settings option for this great ssh client. Never mind the public keys and encryption stuff just the settings and port forward export/import would be a welcome and sorely needed feature on this app. 🙏

Nebur692:
+1

xXAzazelXx:
+1

kbourgeois:
+1

faan11:
+1

nindanaoto:
This issue seems to be fixed by the combination of #1687 and #1688.

Plazzmex:
Hello,

[...]

PR #1687, opened 2025-12-08
Add host configuration export/import via JSON

### LINKED ISSUE #1643: Export/Import server configurations
Due to recent regressions I'm finding it impossible to use the current version of Connect Bot. I'm sure there will be long term improvements, but right now I need it to work despite any previous areas for improvement.

I'd like to revert version, but it's not possible without an uninstall.

I'm concerned that uninstall will wipe all my configuration.

Import/Export of server settings to JSON.

And (ideally)

Import/Export of encrypted file of keys.

nindanaoto:
This resolves #1643. This PR does not include key export/import, as I prefer to use the biometric key and generate a separate key for each device to minimize key leakage risk. I decided to implement this because I noticed I need to reinstall my daily app to check the latest changes, and it sometimes deletes my configs...

This PR allows users to export all host configurations (including port forwards) to a JSON file and import them on another device or after reinstalling. Duplicate hosts are skipped during import based on the nickname.

EDIT: As mentioned in #1089, #527, and #922 are treating the same issue.


kruton (changes requested):

kruton:
I wonder if there is a way to export JSON Schema or create the HostJson from JSON Schema somehow. I worry that over time we will add things that break this import/export.

Maybe a few tests with golden .json files will help avert that possibility.

nindanaoto:
I agree that your concern is worth caring about. Hence, I rewrote the logic to use the database JSON schema to determine the export/import JSON format and to remove all entry-specific, hardcoded import/export logic. Because the database has many unnecessary entries to be exported, this PR determines which should be exported using EXPORT_TABLES and which should not be by EXCLUDED_FIELDS in HostConfigJSON.kt.

app/src/main/java/org/connectbot/data/DatabaseSchema.kt
```diff
@@ -0,0 +1,209 @@
...
+        result
+    }
+
+    /**
+     * Get entity schema by table name.
+     */
+    fun getEntity(tableName: String): EntitySchema? = entities[tableName]
+
+    companion object {
+        private const val SCHEMA_PATH_FORMAT = "org.connectbot.data.ConnectBotDatabase/%d.json"
```
kruton: I don't think this will work because you don't include this in the app…

nindanaoto:
The new commit also fixes the issue with the host count in the importing message. Previously, the number of port forwards was counted as a host.

nindanaoto:
Oh, the CI is complaining about the filtered schema generation. Let me fix that...

nindanaoto:
Now, the problem seems to be resolved. I believe this is now ready for merge.

nindanaoto:
Rebased to main.

nindanaoto:
Something wrong. Let me check...

nindanaoto:
OK, because of the change in the main, I have to update the code to use Hilt's database instead of the old Room database access. Now I believe it is resolved.

nindanaoto:
Oh, CI still fails... Let me check.

nindanaoto:
OK, I also fixed that, after importing the settings, I have to restart the app to see the imported hosts. This might be caused by changing how the app accesses the database.

nindanaoto:
OK, CI is also fixed. I believe this is now ready to merge, again.

kruton:
So one thing I've been planning to do is to have groups of hosts that are synchronized with a shared database. This could be a service that one stands up themselves or maybe it can just read from a Google Drive files. The only way you can do this reliably is to have the Primary Key in each database table as a UUID. Then when you get changes you can match it to existing hosts. Deleting a host would mean just marking it as deleted in the database by putting the UUID in a different table or just enabling a boolean on a field in the Hosts table.

I don't know if this idea would mess up your JSON exports, but I wanted to let you know before I institute this change. The `host.id` will eventually be a UUID instead of an `INTEGER` in the database.

nindanaoto:
If my understanding is correct, such change will not break this JSON feature, at least for catastrophically. If the JSON schema for the table is available, the only thing this feature requires is to select whic table should be saved and which field should be excluded.

nindanaoto:
Rebased to main

nindanaoto:
Since we may also have to export the profile after merging #1697, I'm not expecting this PR to be merged before merging #1697 and applying the required changes to this PR, though the required change might be subtle.

nindanaoto:
Now I have confirmed that we can treat profiles correctly.

kruton (approved):

Issue #1674, opened 2025-12-06
Create profiles for different configurations


kruton:
## Describe the solution you'd like

The current colors screen is not as useful as it could be. I propose changing it to a profiles screen where you can set colors, fonts, and other configurations that have previously been in Preferences. This will allow it to change per-host if desired.

You should also be able to edit the current profile and switch profiles for the current host from the ConsoleScreen.

Since ConnectBot has both "google" and "oss" versions, we can't rely only on Google Play Services for font downloading. There needs to be a local file import option. That should do some checks to make sure the font works (e.g., show a `Text` preview before importing.

Steps:
- [ ] Determine list of preferences that should be moved to profiles
- [ ] Create Profiles database schema
- [ ] Migrate the relevant Preferences to the default Profile
- [ ] Link hosts to Profiles and remove Colors selection (no need to migrate hosts as we are in beta still)
- [ ] Replace navigation to Colors with Profiles
- [ ] Add nav path from Settings screen to Profiles as well (for users that are used to old ConnectBot)
- [ ] Limit Colors palette to the first 16 colors
- [ ] Add downloadable fonts
- [ ] Allow composing fonts like PowerLineExtraSymbols (i.e., ability to add Powerline symbols to any font)

## Additional context

iTerm2 has really good profiles support. Inspirational!

kruton:
@nindanaoto completed this work.

PR #1697, opened 2025-12-09
Add downloadable fonts support with global and per-host settings

### LINKED ISSUE #458: Changing fonts

### LINKED ISSUE #1381: Terminal isn't shown with a fixed width font on Motorola moto g84 5G
## Bug description
The terminal isn't shown with a fixed width font on Motorola moto g84 5G.

## Steps to reproduce

1. Install ConnectBot on a new moto g84 5G.
2. Connect to a host.

## Expected behavior
The terminal is shown with a fixed width font.

## Screenshots

Screenshot_20240312-212546

## Android device

 - Device: Motorola moto g84 5G
 - OS: Android 13
 - ConnectBot Version: 1.9.10

## Server information

 - OS: [e.g., Ubuntu Linux 16.10]
 - SSH Software and Version: [e.g., OpenSSH 7.1]
 - Pubkeys used (if applicable): [e.g., Ed25519]

### LINKED ISSUE #1451: use .ttf font file custom font...

On some devices, the displayed text cannot be aligned.

I want to align the displayed content by specifying the font file.

## Describe alternatives you've considered

none

nindanaoto:
This PR partially resolves #458. As you can see, the current version assumes the local termlib fork because it requires connectbot/termlib#40. 

- Add font selection preference in global settings (Settings screen)
- Add per-host font override option in host editor
- Integrate with termlib's downloadable fonts feature
- Add fontFamily column to Host entity with database migration (v2->v3)
- Use the local termlib module instead of the Maven artifact for font support
- Update ConsoleScreen to apply selected font via rememberTerminalTypeface()

Fixes #1451
EDIT: This might also fix #1381, but the current main's forcing monospace use approach is already enough to resolve that.


kruton:
Thanks for the contribution, but I think font downloading belongs in ConnectBot and not the termlib component.

I also want to rearrange the color schemes to be part of "profiles" one can assign to hosts as described in #1674. If we're adding more configuration to hosts, it will make the host editor too cluttered.

kruton:
I updated the issue with some more notes if you wanted to work on that.

app/src/main/java/org/connectbot/data/ConnectBotDatabase.kt
```diff
@@ -66,12 +71,15 @@ import org.connectbot.data.entity.Pubkey
...
+        Profile::class
     ],
-    version = 2,
+    version = 5,
     exportSchema = true,
     autoMigrations = [
-        AutoMigration(from = 1, to = 2)
+        AutoMigration(from = 1, to = 2),
+        AutoMigration(from = 2, to = 3),
+        AutoMigration(from = 4, to = 5)
```
kruton: This may be annoying to you, but I would expect a PR to only increment the database version once. It will save the millions of ConnectBot installations from doing unnecessary migrations.

app/src/main/java/org/connectbot/ui/screens/console/ConsoleScreen.kt
```diff
@@ -341,22 +344,28 @@ fun ConsoleScreen(
...
-//                            Typeface.CustomFallbackBuilder(
-//                                FontFamily.Builder(
-//                                    Font.Builder(context.assets, "fonts/PowerlineExtraSymbols.ttf").build()
-//                                ).build()
-//                            ).setSystemFallback("monospace").build()
-//                        } else {
-//                            Typeface.MONOSPACE
-//                        }
+                        // Get font: per-host override takes precedence over global default
+                        val globalFontFamily = prefs.getString("fontFamily", "SYSTEM_DEFAULT") ?: "SYSTEM_DEFAULT"
```
kruton: This should be reading from the profile. I'd also like to keep the fallback building done in the removed comment. The fallback font should be `Typeface.MONOSPACE` before going to the system default. Typically the system fallback font is not a monospace font.

What I'd like to have happen is that a user can compose two fonts in a profile. This is because things like [Powerline Extra Symbols](https://github.com/ryanoasis/powerline-extra-symbols) can usually be downloaded separately from an entire font. Then the `Typeface.CustomFallbackBuilder` might look like:

    Powerline Extra Symbols → Fira Code → monospace

The UI for configuring that might get extremely complicated, so I'm happy with either only allowing two fonts or a checkbox that says something like "Use Powerline Extra Symbols"

app/src/main/java/org/connectbot/util/TerminalFontCompose.kt
```diff
@@ -0,0 +1,234 @@
...
+        Log.d(TAG, "Loading font from provider: $googleFontName")
+        val loadedTypeface = suspendCancellableCoroutine { continuation ->
+            fontProvider.loadFontByName(googleFontName) { result ->
+                if (continuation.isActive) {
+                    Log.d(TAG, "Font loaded: $googleFontName, typeface: $result")
+                    continuation.resume(result)
+                }
+            }
+        }
+        typeface = loadedTypeface
```
kruton: I'd like this to use the `CustomFallbackBuilder` as I mentioned in another comment.

app/src/main/java/org/connectbot/di/DatabaseModule.kt
```diff
@@ -39,7 +39,9 @@ object DatabaseModule {
             context,
             ConnectBotDatabase::class.java,
             DATABASE_NAME
-        ).build()
+        )
+            .addMigrations(ConnectBotDatabase.MIGRATION_3_4)
```
kruton: Strange this has to be added here. It might need to be added to any `TestDatabaseModule` as well.

app/src/main/java/org/connectbot/util/TerminalFontProvider.kt
```diff
@@ -0,0 +1,236 @@
...
+import androidx.core.provider.FontsContractCompat
+import org.connectbot.R
+import java.util.concurrent.ConcurrentHashMap
+
+/**
+ * Provider for loading terminal fonts via Google Fonts API.
+ * Fonts are cached after loading for performance.
+ * Supports both preset fonts (TerminalFont enum) and custom font names.
+ */
+class TerminalFontProvider(private val context: Context) {
```
kruton: I'm suspicious that this class doesn't use `suspend fun` and also doesn't reference the `Dispatchers.IO` to touch the disk. Is this causing some jank in the UI? Currently the `ConsoleScreen.kt` is disabling `StrictMode` which might be hiding some problems in this code.

nindanaoto:
As you requested, I partially implemented your issue. At least, composable font is missing. If you really need, may be I can split this to multiple PRs because this PR  now includes 6k+ lines changes. 

Major Features                                                                                                 
                                                                                                                 
  Profile System                                                                                                 
  - New Profile entity that bundles terminal settings: font family, font size, encoding, DEL key behavior,       
  terminal emulation, and color scheme                                                                           
  - Profile list screen with create, duplicate, and delete actions                                               
  - Profile editor screen for modifying all bundled settings                                                     
  - Hosts can optionally select a profile to inherit settings from                                               
  - Default profile setting in app Settings (applied to new hosts automatically)                                 
  - Built-in "Default" profile that cannot be deleted                                                            
  - When a profile is selected for a host, the profile-controlled settings are disabled in the host editor       
                                                                                                                 
  Downloadable Fonts Support                                                                                     
  - Integration with Google Fonts API for monospace fonts (JetBrains Mono, Fira Code, Source Code Pro, etc.)     
  - Font caching and preloading for smoother user experience                                                     
  - Custom font support - users can add any Google Font by name                                                  
  - Per-host font override in addition to global setting                                                         
  - Snackbar notification when font loading fails                                                                
                                                                                                                 
  Local Font Support                                                                                             
  - Import TTF/OTF/TTC fonts from device storage                                                                 
  - Local fonts work in OSS builds without Google Play Services                                                  
  - Downloadable fonts hidden in OSS builds (require Play Services)                                              
                                                                                                                 
  Custom Terminal Types                                                                                          
  - Users can add/remove custom TERM values in Settings                                                          
  - Expanded preset terminal emulation options (vt102, vt220, linux, dumb)                                       
                                                                                                                 
  Bug Fixes                                                                                                      
                                                                                                                                                      
  - Fix black screen when rapidly tapping back button (navigation race condition)                                                                          
                                                                                                                 
  Database Changes                                                                                               
                                                                                                                 
  - Migration from v2 to v5 with new Profile table and fontFamily/profileId fields on Host

nindanaoto:
Oh, I forgot to mentione that the terminal color schema is still bit buggy, the initial SSH connection messages is still not followong the schemea. I'm currently thinking it as a termlib's bug. Although it is still buggy, I believe that having the capability to change the schema is better than nothning.

nindanaoto:
Oh, I did not notice your comments on my code because it was before I declared I would resolve much more from #1674. Let me check your comments and fixes.

app/src/main/java/org/connectbot/data/ConnectBotDatabase.kt
```diff
@@ -82,4 +90,43 @@ abstract class ConnectBotDatabase : RoomDatabase() {
...
+        val MIGRATION_3_4 = object : Migration(3, 4) {
+            override fun migrate(db: SupportSQLiteDatabase) {
+                // Create profiles table
+                // Note: No foreign key to color_schemes because built-in color schemes use negative IDs
+                // and are virtual (not stored in the database). Only custom schemes have positive IDs.
+                db.execSQL("""
+                    CREATE TABLE IF NOT EXISTS `profiles` (
+                        `id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
+                        `name` TEXT NOT NULL,
+                        `is_built_in` INTEGER NOT NULL DEFAULT 0,
```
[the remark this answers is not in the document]
nindanaoto: Now using transient @Ignore field, isBuiltIn is now computed from id == DEFAULT_PROFILE_ID. I believe this is resolved.

app/src/main/java/org/connectbot/data/entity/Host.kt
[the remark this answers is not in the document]
nindanaoto: Done. Removed fontSize, colorSchemeId, delKey, encoding, and fontFamily from the Host entity. The migration now creates profiles from unique terminal settings combinations using a ProfileSettingsKey data class to group hosts with identical settings into shared profiles.

app/src/main/java/org/connectbot/ui/navigation/NavGraph.kt
```diff
@@ -36,10 +37,26 @@ import org.connectbot.ui.screens.hints.HintsScreen
...
 import org.connectbot.ui.screens.pubkeyeditor.PubkeyEditorScreen
 import org.connectbot.ui.screens.pubkeylist.PubkeyListScreen
 import org.connectbot.ui.screens.settings.SettingsScreen
 
+/**
+ * Safely pops the back stack, preventing double navigation when the user rapidly taps
+ * the back button. This checks if the current destination's lifecycle state is RESUMED
+ * before allowing the navigation to proceed.
+ */
+fun NavHostController.safePopBackStack(): Boolean {
```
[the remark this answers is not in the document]
nindanaoto: Updated to use the suggested idiomatic pattern with a lifecycleIsResumed() extension function on NavBackStackEntry?.

app/src/main/java/org/connectbot/ui/screens/console/ConsoleScreen.kt
nindanaoto: Partially fixed. ConsoleScreen now reads font only from the profile via bridge.fontFamily. Removed the host-level font override and global preference fallback logic. 
Added scaffolding for buildTypefaceWithFallback() in TerminalFontCompose. The Full CustomFallbackBuilder implementation is limited because the Font.Builder API doesn't accept an existing Typeface directly - it requires raw font data (from assets or resources). The current implementation returns the loaded typeface, and Android's text system provides a natural fallback for missing glyphs. If we want full control over the fallback chain, we'd need to load fonts from files/assets rather than via the Google Fonts API. 

For a locally downloaded font, we can do the composable font, I guess, but it's still not implemented.

kruton:
Great work! I will review this thoroughly soon.

おつかるさまです 🙇🏼‍♂️

app/src/main/java/org/connectbot/ui/screens/console/ConsoleScreen.kt
nindanaoto: Although I believe so, adding the functionality might not break other functionality, so now working on this. If there are significant obstacles, may be I will give up, but still I do not observe such things.

nindanaoto: Well, I found that CustomFallbackBuilder does not work for such character-level replacement. So, this functionality is not implementable at the connectbot level. We need a patch for termlib, but I'm not going to include that in this PR. In short, I give up this time......

nindanaoto:
This PR is now ready to merge if your further review is fine.

kruton:
This still has quite a few bugs, but I can fix them if you're not interested in working further on this.

Here are the things I noticed:
- When changing a profile's attributes, a connected host doesn't update
- When changing the host's profile, the connected host doesn't update
  - [ ] This is probably because it's not using Kotlin Flows?
- When first installed there is no setting, but if you name one "Default" it says it's built-in
  - [ ] A default profile should be created on migration or database initialization (named "Default" or whatever the local language is) and shouldn't be treated differently than other profiles
- You can edit the "Default (built-in)" profile but it doesn't let you save it
  - [ ] There shouldn't be any special profiles

nindanaoto:
Let me fix that.

kruton:
I created a small merge conflict for you. I changed the colorSchemeId default to -1L. Your change deletes the lines anyway, so it can be overcome by deleting the lines.

nindanaoto:
I believe they are now fixed.

nindanaoto:
I'm currently expecting the main branch will version up termlib first, then I will rebase again and manually tests color schema related things.

kruton:
I have been testing this change, but I haven't yet concluded. You could probably guess from the termlib change.

My response time is going to slow down quite a bit in the next week or so. I'll be back after Christmas. But before that I need to push a schema update to fix a migration issue that someone is having, so it will cause a conflict with your `3.json` file. You should be able to change the database revision to 4 to get past the conflict which will create a `4.json` again.

nindanaoto:
I got your situation. Since this is an OSS and a large PR, it is acceptable to take a long time to merge. Then I will version up a termlib on this PR side to proceed with further integration testing on my side.

nindanaoto:
Oh, I found that termlib is now up to date. I rebased this branch, so let me test it......

nindanaoto:
I confirm that the color schema for the terminal seems to be working correctly now.

PR #1698, opened 2025-12-09
Fix foreign key constraint error during database migration

### LINKED ISSUE #1675: DB migration failures with latest beta, unable to launch
## Bug description

ConnectBot fails immediately on launch since update to Dec 6 beta.

## Steps to reproduce

1. Launch ConnectBot

## Expected behavior

App loads

## Android device

 - Device: Samsung Galaxy A53
 - OS: Android 16 / ONE UI 8.0
 - ConnectBot Version: git-1.9.13-108

## Server information

N/A

nindanaoto:
This PR possibly fixes #1675. Because I could not come up with a way to reproduce the error, this is solely based on a guess from the error message, so I can say only 'possibly' fix.   

Room auto-generates IDs when inserting entities with autoGenerate=true, but the migration code was using the original IDs for foreign key references. This caused SQLITE_CONSTRAINT_FOREIGNKEY errors when inserting ColorPalette entries that referenced the old ColorScheme IDs.

Add proper ID mapping for ColorScheme, Pubkey, and Host entities during migration, and remap all foreign key references (ColorPalette.schemeId, Host.pubkeyId, Host.colorSchemeId) to use the new database-assigned IDs.

### DIFF
--- a/app/src/main/java/org/connectbot/data/migration/DatabaseMigrator.kt
+++ b/app/src/main/java/org/connectbot/data/migration/DatabaseMigrator.kt
@@ -491,26 +491,59 @@ class DatabaseMigrator @Inject constructor(
         // Wrap all writes in a single transaction for atomicity
         // If any write fails, all writes are rolled back
         roomDatabase.withTransaction {
-            // Insert color schemes first (referenced by hosts)
+            // Insert color schemes first (referenced by hosts and palettes)
+            // Create ID mapping since Room auto-generates IDs
+            val colorSchemeIdMap = mutableMapOf<Long, Long>()
             data.colorSchemes.forEach { scheme ->
-                roomDatabase.colorSchemeDao().insert(scheme)
+                val oldId = scheme.id
+                val newId = roomDatabase.colorSchemeDao().insert(scheme)
+                colorSchemeIdMap[oldId] = newId
             }
 
-            // Insert color palettes
+            // Insert color palettes with remapped scheme IDs
             data.colorPalettes.forEach { palette ->
-                roomDatabase.colorSchemeDao().insertColor(palette)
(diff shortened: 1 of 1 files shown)


kruton:
I'll have to check whether default colorScheme is actually 1.

ferdnyc:
@kruton 

> I'll have to check whether default colorScheme is actually 1.

Based on my output in #1675 (as someone who has never set/customized anything regarding color schemes, so anything stored should be set to defaults), it seems like it might be `0`?

```text
Step 2: Validating data
WARNING: Found 4 host(s) referencing non-existent SSH keys. References will be cleared.
WARNING: Found 6 host(s) referencing non-existent color schemes. Will use default color scheme.
WARNING: Found 2 color palette(s) referencing 1 non-existent color scheme(s): 0. Will synthesize missing schemes.
Step 3: Transforming data to Room entities
Synthesizing missing ColorScheme with ID 0 for orphaned palette entries
Clearing invalid pubkey reference (ID 1) for host '*****'
Resetting invalid color scheme reference (ID 1) to default for host '*****'
Clearing invalid pubkey reference (ID 1) for host '*****'
Resetting invalid color scheme reference (ID 1) to default for host '*****'
Clearing invalid pubkey reference (ID 2) for host '*****'
Resetting invalid color scheme reference (ID 1) to default for host '*****'
Resetting invalid color scheme reference (ID 1) to default for host '*****'
Clearing invalid pubkey reference (ID 1) for host '*****'
Resetting invalid color scheme reference (ID 1) to default for host '*****'
Resetting invalid color scheme reference (ID 1) to default for host 'shell'
Recovery: Cleared invalid pubkey references from 4 host(s)
Recovery: Synthesized 1 missing color scheme(s) for orphaned palette entries
```

(**Edit:** Grabbed the right output this time.)

PR #1719, opened 2025-12-12
Multiple sessions

### LINKED ISSUE #1522: Multiple connections to same host?
When using other terminal apps I often have multiple terminals / tabs open to my home server to have multiple activities on the go that I can switch between via the application tabs or swipe between with ConnectBot. 
This feels quicker than screen/tmux sessions.

However I'm unable to find a way to start a new session to the host definition for my local server, is this possible? 
Even if I manually define a new host with the same connect string, when I try to open it, it instead jumps to the existing open session?

Thanks!

nindanaoto:
This resolves #1522. I decided to implement this because a similar feature is available in Termius.

Current behavior:
- **Opening sessions**: Tapping the host opens a single new session or connects to an existing one. Long-pressing the host displays an option to open a new session rather than connecting to the existing one. Alternatively, you can add a new session via the menu in the terminal view.
- **Deleting sessions**: Closing the connection from the host menu closes all sessions. Closing the connection from the terminal view closes only the current session.
- **Switching sessions**: The only way to switch sessions is via the menu in the terminal view.

I initially considered using a horizontal swipe to switch sessions. However, that gesture seems to be consumed by termlib, and I want to avoid including changes to termlib in this PR.

### DIFF
--- a/app/src/androidTest/java/org/connectbot/service/TerminalBridgeWriteSerializationTest.kt
+++ b/app/src/androidTest/java/org/connectbot/service/TerminalBridgeWriteSerializationTest.kt
@@ -182,7 +182,7 @@ class TerminalBridgeWriteSerializationTest {
                 main = Dispatchers.Main,
             )
 
-            val bridge = TerminalBridge(manager, host, dispatchers)
+            val bridge = TerminalBridge(manager, host, sessionId = 1L, dispatchers = dispatchers)
             try {
                 val transport = LossyNonThreadSafeTransport()
                 bridge.transport = transport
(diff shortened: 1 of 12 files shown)


kruton:
I'm working on restructuring the app to be an adaptive layout, but the pull requests you've been opening might be conflicting when it's done.

nindanaoto:
I'm happy to keep PRs rebased on main. Don't worry about conflicts when introducing drastic changes to main ;)

igor-ramazanov:
Reading the code, can't figure out if it uses something like: https://man.archlinux.org/man/ssh_config.5#ControlMaster ?

Looks like it doesn't reuse the same network connection against same host? On PC it significantly improves the UX: no need to auth again and it's a lot faster at creating new connections.

nindanaoto:
Yes, currently using a new session. I've never noticed that option, but since it is an option, I hope to keep current behavior as a simplest PR, and possibly add that as an another PR. Thank you for your suggestion.

igor-ramazanov:
Agree, something for the future

nindanaoto:
I noticed that this is currently not compatible with the terminal swotch button. I will fix that.

Issue #1419, opened 2024-06-18
Can't add new host in portrait mode


pegranka:
Bug description
Cannot add new host when in portrait mode on Samsung Galaxy Tab S9 FE

## Steps to reproduce
Click + to add host
Shows screen with Protocol ssh and a computer icon

## Expected behavior
Screen with fields for adding host details

## Screenshots
See screenshot 

## Android device
 - Device: Samsung Tab S9 FE, SM-X510
 - OS: Android 14, One UI 6.1
 - ConnectBot Version: 1.9.10

## Server information
N/A

## Additional context
Works correctly in landscape mode

![Screenshot_20240618_095931_ConnectBot](https://github.com/connectbot/connectbot/assets/26248322/f0d9ee19-f5d9-4912-8166-10d2d3429630)

pegranka:
Editing an existing host works normally

pegranka:
Also works if add host in portrait (shows screen in above screenshot), then rotate tablet to landscape

lianergoist:
On my Lenovo Tab M10 tablet I can view list of hosts in both portrait and landscape mode, but the terminal window is always portrait - even when the tablet is in landscape mode. (2000x1200)

pegranka:
Have a look at Rotation mode in the app settings

larsskj:
That happens for me as well using a Lenovo tablet.

kruton:
This originally was opened against v1.9 series, but the new beta Jetpack Compose v1.10 version removed this rotation setting. I can add it back if there are still buggy Android devices that don't work correctly.

pegranka:
Version 1.9.13 works in portrait mode on SM-X510. This version of connectbot still allows screen rotation. Landscape more is handy occasionally when want to view long lines (eg in a log file) without wrapping.

kruton:
Is there a reason you can't just rotate the screen to do it? Some devices used to have broken rotation sensors (it would be inverted, etc)

pegranka:
Rotating screen works fine on Tab S9. I leave Connectbot rotation mode setting on Auto. v1.10 not yet in app store.

kruton:
If you're adventurous, you can enable the beta in Play Store to see the new design.

pegranka:
Hi, 

[...]

PR #1791, opened 2025-12-25
Minimal Integration Test

### LINKED ISSUE #639: Add SSH integration tests
To prevent issues with proguard, et al., (like #637) we need integration tests. I have had success with [testcontainers](https://www.testcontainers.org/) in the SSH library project.

nindanaoto:
This PR partially fixes #639. Only adds the SSH connection test using the Android emulator and Docker. If this direction for the test implementation is fine, we can move on to much more complex test cases.


kruton:
I wrote the integration test for `sshlib` using testcontainers, so I'm not entirely sure this is useful. You can look at the [integration test](https://github.com/connectbot/sshlib/blob/main/src/test/java/com/trilead/ssh2/OpenSSHCompatibilityTest.java) and [docker files](https://github.com/connectbot/sshlib/tree/main/src/test/resources/openssh-server).

I have a feeling it will make the already long test time even longer. I'd like to keep the CI under 15 minutes and they're already over.

nindanaoto:
My initial motivation was to identify cases like #1419 and #586 by providing a full-fledged testing environment, with the connection test serving as a minimal example. I agree that this alone is not useful enough. 
However, I also agree that the testing time is too long. I will keep the current testing scheme without the Android emulator to resolve the initial motivations...

PR #1792, opened 2025-12-25
Re-encrypt imported encrypted keys with user's password

### LINKED ISSUE #1789: Cannot forget password for a key
## Bug description
Once the password for a key has been entered, connecting to a host never asks for key. Should be an option to request password each time connect

## Steps to reproduce
1. Click Manage pubkeys 
2. Import an existing key 
2. Enter password
3. Connect to a host
4. Never asks for password on future connections 
5. In Manage pubkeys clicking a key toggles green ring, but has no other effect

## Expected behavior
Toggling been ring of should forget password.
There should be an option to remember password for a day time.

## Android device
 - Device: Samsung S9 Fe
 - OS: Android 16
 - ConnectBot Version: git-v1.9.13-242

nindanaoto:
Fix #1789.
Previously, when importing an encrypted SSH key, the key was decrypted but stored unencrypted in the database. Now the key is re-encrypted using the internal encryption format with the user's original password.

### DIFF
--- a/app/src/main/java/org/connectbot/ui/screens/pubkeylist/PubkeyListViewModel.kt
+++ b/app/src/main/java/org/connectbot/ui/screens/pubkeylist/PubkeyListViewModel.kt
@@ -758,15 +758,18 @@ class PubkeyListViewModel @Inject constructor(
             val kp = PEMDecoder.decode(keyString.toCharArray(), password)
             val algorithm = convertAlgorithmName(kp.private.algorithm)
 
+            // Re-encrypt the private key with the user's password for secure storage
+            val encryptedPrivateKey = PubkeyUtils.getEncodedPrivate(kp.private, password)
+
             return Pubkey(
                 id = 0,
                 nickname = nickname,
                 type = algorithm,
-                encrypted = false,  // Store decrypted in internal format
+                encrypted = true,  // Key is now encrypted with user's password
                 startup = false,
                 confirmation = false,
                 createdDate = System.currentTimeMillis(),
-                privateKey = kp.private.encoded,
+                privateKey = encryptedPrivateKey,
                 publicKey = kp.public.encoded
             )
         } catch (e: Exception) {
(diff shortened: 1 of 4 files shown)


kruton:
It should probably be an import option. Something like this:

Screenshot 2025-12-25 at 11 06 12 AM

app/src/main/res/values/strings.xml
```diff
@@ -126,6 +126,10 @@
 	<string name="pubkey_import_encrypted_title">"Import encrypted key"</string>
 	<string name="pubkey_import_encrypted_message">"Enter password to decrypt key '%1$s' (%2$s)"</string>
 	<string name="pubkey_import_button">"Import"</string>
+	<string name="pubkey_import_encrypt_key">"Encrypt key for storage"</string>
```
kruton: The comment before each line should contain a description for translators to know how to translate it.

app/src/main/java/org/connectbot/ui/screens/pubkeylist/PubkeyListViewModel.kt
```diff
@@ -750,23 +753,36 @@ class PubkeyListViewModel @Inject constructor(
         _uiState.update { it.copy(pendingImport = null) }
     }
 
-    private fun decryptAndImportKey(keyData: ByteArray, nickname: String, password: String): Pubkey? {
+    private fun decryptAndImportKey(
```
kruton: Can you add some tests for this?

kruton (changes requested):

app/src/main/res/values/strings.xml
nindanaoto: Added comments.

nindanaoto: Added comments.

nindanaoto:
I believe now this PR implements what you requested;)

kruton (approved):

kruton:
Thanks, looks good.

Issue #1044, opened 2021-11-19
How to transfer files by using Connectbot?


Amazingkenneth:
Hi,

I tried to search for methods such as SCP/SFTP that can help me transfer files between devices. However, it seems that there is no way to transfer files using _Connectbot_ or client that only support ssh.

So how can I transfer files, or maybe it's just a unfinished feature?

avipars:
Would love this as a feature. 

Till its implemented, doesn't termux allow this?

PR #1810, opened 2025-12-29
SFTP Client

### LINKED ISSUE #52: scp/sftp functionality
it would be nice, if files could be shared via scp/sftp

### LINKED ISSUE #1264: Local shell & SSH: Storage access and possible implementations & alternatives.
Can't access local storage when using local shell, can't use PWD, LS or related commands

An option to enable access of all directories to a shell, which would allow for use with Shizuku(Local ADB SHELL) or SFTP.

## Describe alternatives you've considered
An option to add visible directories to a shell, so it only sees a folder you selected as its [/]
, also I think you could make ConnectBot create a org.connectbot directory at Android/Data/org.connectbot/ so the app can access it without additional permissions.

nindanaoto:
This resolves #52, #1044, and partially resolves #1264. As the name suggests, this PR adds simple SFTP client functionality. To use this SFTP client, the user should open the three-dot menu for the desired host and select the new SFTP menu. I have confirmed that I can download and send some data between my phone and the host.


skyerjoe:
What about scp ?

Use case: Some openwrt router demand  this 

Regards

nindanaoto:
I'm not planning to add scp to this PR. Adding SFTP is already enough size for one PR. I know that OpenWRT is using dropbear and it only supports scp, but I believe that adding scp should be another PR. If you really need that, I recommend you to open a new issue to track that issue.

nindanaoto:
I'm not planning to add scp to this PR. Adding SFTP is already enough size for one PR. I know that OpenWRT is using dropbear and it only supports scp, but I believe that adding scp should be another PR. If you really need that, I recommend you to open a new issue to track that issue.

nindanaoto:
I found that I regulary requires this feature in my usecase. Though I can use my local build for a while, if we can get this merged in early windows, I'm happy with that.

Fei1Yang:
> What about scp ?
> 
> Use case: Some openwrt router demand this
> 
> Regards

Note that if you install `openssh-sftp-server` (it's very small and don't bring the whole OpenSSH in), Dropbear can also do SFTP.

nindanaoto:
I noticed ProxyJump requires multiple biometric authentication for the same key. I will fix that.

nindanaoto:
The above ProxyJump issue seems to be fixed.

Issue #1813, opened 2025-12-29
"Copy public key" doesn't work on Ed25519 keys on Pixel 10 Pro XL


pkoepke:
## Bug description
When you try to copy the public key for a Ed25519 key pair, nothing happens on a Pixel 10 Pro XL (works fine on a Pixel 8a, details of both below).

## Steps to reproduce
1. Generate a Ed25519 key pair.
2. In the list of keys, long-press on the Ed25519 key and tap "Copy public key". Nothing happens

Copying RSA public keys works, and copying Ed25519 private keys works. I tried copying the Ed25519 public key and pasting it just in case the problem is with the toast, but this confirmed that the public key really is not copied.

## Expected behavior
The key should be copied and a toast with the copied text appears.

## Android device
Device where it does not work:
 - Device: Google Pixel 10 Pro XL
 - OS: stock Android 16 build number BD3A.251105.010.E1
 - ConnectBot Version: 1.9.13

Device where copying the public Ed25519 key works:
 - Device: Google Pixel 8a
 - OS: stock Android 16 build number BP3A.251105.015
 - ConnectBot Version: 1.9.13

nindanaoto:
Whmmm, it is strange behavior. I'm not sure why it depends on the device. I guess that you are using the release version. Could you also try the beta version? I rewrote the key export-related logic in the beta, which may resolve the issue as a side effect.

kruton:
It might be Google Play Services being different versions. @pkoepke can you check the version of it on each device?

pkoepke:
Google Play Services version:

Pixel 10: version 25.47.61 (260400-839048624)

Pixel 8a: version 25.47.30 (260400-833691957)

pkoepke:
I can try the beta version, what's the best way to get it? It won't wipe out my saved hosts or keys, right?

kruton:
Some have said it erased some of their pubkeys. I haven't been able to reproduce it, though.

pkoepke:
That's fine, if my key gets erased it wouldn't be hard to make a new one and distribute it to my computers. Which release should I use to test this?

nindanaoto:
FYI: #1623 is the mentioned issue.

nindanaoto:
You can install the beta version by joining the google play's beta tester program from the connectbot's store page or install from the latest pre release build's apk. https://github.com/connectbot/connectbot/releases/tag/git-v1.9.13-277-gcb2f303a

pkoepke:
[...]

PR #1817, opened 2025-12-31
Fix Migrate from legacy Again

nindanaoto:
This PR fixes the case that happened in the migration to the beta process in #1813. I found that I embedded a bug in #1807. 

Summary                                                                                                                                                                                                                             
                                                                                                                                                                                                                                      
  - Fix race condition in DatabaseMigrator.isMigrationNeeded() that caused legacy SSH keys and hosts to not be migrated when upgrading from older ConnectBot versions                                                                 
  - Add a regression test to prevent future occurrences                                                                                                                                                                                 
                                                                                                                                                                                                                                      
  Test plan                                                                                                                                                                                                                           
                                                                                                                                                                                                                                      
  - Run LegacyMigrationRaceConditionTest - verifies the fix works                                                                                                                                                                     
  - Run full unit test suite - all tests pass                                                                                                                                                                                         
  - Manual test: Install old ConnectBot, import SSH keys, upgrade to this version, verify keys are migrated                                                                                                                           
                                                                                                                                                                                                                                      
  Root Cause Analysis                                                                                                                                                                                                                 
                                                                                                                                                                                                                                      
  When users upgraded from legacy ConnectBot (using SQLite database hosts and pubkeys) to the new beta (using Room database), their SSH keys disappeared.                                                                            
                                                                                                                                                                                                                                      
  The bug was in DatabaseMigrator.isMigrationNeeded():                                                                                                                                                                                
                                                                                                                                                                                                                                      
  1. The method checked if the Room database had any data by querying profileDao().getAll()                                                                                                                                               
  2. This query triggered lazy initialization of Room                                                                                                                                                                                 
  3. Room's onCreate callback in DatabaseModule inserted a default profile                                                                                                                                                            
  4. The check then saw data in the profiles table and concluded "Room has data, skip migration."                                                                                                                                      
  5. Legacy pubkeys and hosts were never migrated                                                                                                                                                                                     
                                                                                                                                                                                                                                      
  Fix                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                      
  Check if the Room database file exists before accessing Room. This prevents the lazy initialization that caused the race condition:                                                                                                 
                                                                                                                                                                                                                                      
  val roomDbFile = context.getDatabasePath(ROOM_DATABASE_NAME)                                                                                                                                                                        
  if (!roomDbFile.exists()) {                                                                                                                                                                                                         
      // Migration needed - don't trigger Room initialization                                                                                                                                                                         
      return true                                                                                                                                                                                                                     
  }                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                      
  Also removed the profiles table check since DatabaseModule.onCreate() always creates a default profile, making it unreliable for detecting user data.

### DIFF
--- a/app/src/main/java/org/connectbot/data/migration/DatabaseMigrator.kt
+++ b/app/src/main/java/org/connectbot/data/migration/DatabaseMigrator.kt
@@ -130,6 +130,27 @@ class DatabaseMigrator @Inject constructor(
             return@withContext false
         }
 
+        // Check if legacy databases have any actual data to migrate.
+        // If legacy DBs exist but are empty, skip migration to avoid inserting
+        // a duplicate default profile (fixes #1806).
+        // Wrap in try-catch to handle malformed legacy databases gracefully.
+        val legacyHostsHaveData = legacyHostsExists && try {
+            legacyHostReader.readHosts().isNotEmpty()
+        } catch (e: Exception) {
+            logDebug("Failed to read legacy hosts database: ${e.message}")
+            false
+        }
+        val legacyPubkeysHaveData = legacyPubkeysExists && try {
+            legacyPubkeyReader.readPubkeys().isNotEmpty()
+        } catch (e: Exception) {
+            logDebug("Failed to read legacy pubkeys database: ${e.message}")
+            false
+        }
+        if (!legacyHostsHaveData && !legacyPubkeysHaveData) {
+            logDebug("Legacy databases exist but are empty or unreadable, no migration needed")
+            return@withContext false
+        }
+
         // IMPORTANT: Check if Room database FILE exists BEFORE accessing Room.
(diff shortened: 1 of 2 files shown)


kruton (approved):
Great, this is exactly the situation I was thinking about today.

kruton:
Oh, need to update the other tests:

```
org.connectbot.data.migration.ProfileConflictMigrationTest > migrationSkippedWhenRoomHasProfiles[emulator-5554 - 10] FAILED 
	org.junit.ComparisonFailure: [Migration should NOT be needed because Room already has profiles (fix for #1806)]
	Expecting value to be false but was true expected:<[fals]e> but was:<[tru]e>
```

nindanaoto:
Now CI seems to be fixed.

PR #1837, opened 2026-01-13
Add OpenSSH config format import/export

### LINKED ISSUE #38: Import[/export] settings from[/to] ssh_config
I have a collection of hosts. It's very convenient to have aliases for them. I'm talking about standard feature that `.ssh/config` provides.

I think that it would be nice to reuse this file to have all hosts configuration imported just by importing `.ssh/config`. I would like to have export feature too, but it's much less important than the import.

nindanaoto:
Fix #38. 
Adds the ability to import and export host settings in OpenSSH config format (~/.ssh/config style), complementing the existing JSON format.

Export:
- Only SSH hosts exported (telnet/local skipped)
- Maps Host fields to SSH directives (HostName, User, Port, etc.)
- Resolves pubkey IDs to IdentityFile paths
- Resolves jump host IDs to ProxyJump references
- Includes port forwards as LocalForward/RemoteForward/DynamicForward
- Adds ConnectBot-specific settings as comments

Import:
- Parses standard SSH config format
- Uses Default profile for all imported hosts
- Shows warnings for unsupported directives (ProxyCommand, Match, Include, ServerAliveInterval, etc.)
- Handles duplicate detection and ProxyJump resolution

### DIFF
--- a/app/src/main/java/org/connectbot/ui/screens/hostlist/HostListScreen.kt
+++ b/app/src/main/java/org/connectbot/ui/screens/hostlist/HostListScreen.kt
@@ -94,9 +94,7 @@ import org.connectbot.ui.theme.ConnectBotTheme
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
 fun HostListScreen(
-    makingShortcut: Boolean = false,
     onNavigateToConsole: (Host) -> Unit,
-    onSelectShortcut: (Host) -> Unit = {},
     onNavigateToEditHost: (Host?) -> Unit,
     onNavigateToSettings: () -> Unit,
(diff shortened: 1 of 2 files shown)


kruton:
I'm not sure if this belongs in ConnectBot. There could be an external utility that takes the exported JSON configs and builds the OpenSSH configs or vice versa.

nindanaoto:
As you suggested, I've tried to write such a script, and it seems to be working, so I'm closing this PR. 
https://github.com/nindanaoto/SSHConfigConverter4ConnectBot
I still believe direct import/export functionality would be useful. However, since the Room database-related functionality, including import/export, is relatively new, introducing a new feature for shallow wrapping does not seem like the right approach.

Issue #1846, opened 2026-01-15
Virtual keyboard is no longer persistent within an open connection


DeprecatedHardware:
## Bug description
When a connection is made within connectIbot, the virtual keyboard is no longer automatically displayed. I  consider this a regression within the application as this is not noted in the change log.

## Steps to reproduce

1. Open a connection
2. the connection opens in the app
3. the virtual keyboard is no longer displayed at the bottom of the screen as it was in the legacy application
4. you can click on the tiny keyboard logo to display the keyboard as long as you had enabled 'special keys always available' in the keyboard settings section within the settings configuration menu, otherwise there is no keyboard visible.

## Expected behavior
Once a session is established, the virtual keyboard should be displayed as in the legacy connectbot application., or there should be a per session or global option to automatically display the keyboard to assist with current workflows 

## Android device
Pixel 6A
Android 16
Connectbot git-v1.9.13-304-ga77058fa

nindanaoto:
I reproduced this issue if the key is already unlocked.

DeprecatedHardware:
Confirmed. I use keys to connect to my host and with the key enabled in the connection profile, there is no virtual keyboard presented. If I temporarily disable using a key the password prompt comes up when connecting, with the virtual keyboard and remains on once the password has been entered correctly.

DeprecatedHardware:
@nindanaoto 

it looks like this now presents the virtual keyboard which is good news. with this version with the strange behavior in  https://github.com/connectbot/connectbot/issues/1847 present.

nindanaoto:
I believe that this issue is resolved.

PR #1851, opened 2026-01-18
Show software keyboard when SSH session connects

nindanaoto:
Fixes GitHub issue #1846 where the virtual keyboard no longer automatically displays when establishing an SSH connection.

Two issues were identified and fixed:

1. Added a LaunchedEffect that monitors the session state and shows the software keyboard when the session becomes open (if no hardware keyboard is connected).

2. Fixed the IME sync logic that was canceling keyboard show requests before the keyboard had a chance to appear. The sync now only triggers after the keyboard has been visible at least once, which prevents it from interfering with the initial show.

Also fixes pre-existing ktlint compose rule violations:
- Added modifier parameter to ConsoleScreen
- Added viewModel as default parameter per compose best practices
- Used rememberUpdatedState for navigation callback in LaunchedEffect

### DIFF
--- a/app/src/main/java/org/connectbot/ui/screens/console/ConsoleScreen.kt
+++ b/app/src/main/java/org/connectbot/ui/screens/console/ConsoleScreen.kt
@@ -271,12 +271,11 @@ fun ConsoleScreen(
     val snackbarHostState = remember { SnackbarHostState() }
 
     // Show software keyboard when session becomes open (if no hardware keyboard)
-    var previousSessionOpen by remember { mutableStateOf(false) }
-    LaunchedEffect(sessionOpen, hasHardwareKeyboard) {
-        if (!previousSessionOpen && sessionOpen && !hasHardwareKeyboard) {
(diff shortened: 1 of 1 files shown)


kruton:
I think there's a way to see if the current window is at the foreground which seems like a more direct way of checking this. I haven't looked at it closely, but saw it mentioned in multi window support documentation.

nindanaoto:
I found that the previous logic sometimes fails. I'm not sure how to reproduce it consistently, but fixing the logic with a new commit seems to be the solution.

I tried to apply your idea to this behavior. However, since the problem happens when we connect to the server, the app is already focused. Are you considering a corner case I'm not currently aware of?

kruton:
> I found that the previous logic sometimes fails. I'm not sure how to reproduce it consistently, but fixing the logic with a new commit seems to be the solution.
> 
> I tried to apply your idea to this behavior. However, since the problem happens when we connect to the server, the app is already focused. Are you considering a corner case I'm not currently aware of?

Not sure. We can try this first.

PR #1909, opened 2026-01-30
Add optional biometric/PIN authentication on app launch

### LINKED ISSUE #1754: Feature Request: User verification/authentication on app launch
It would be really nice, for security reasons to have the ability to enable fingerprint or pin authentication when the app launches for the first time.

nindanaoto:
Fix #1754. 
Gate the app UI behind a BiometricPrompt authentication screen when the "Authentication on launch" preference is enabled. Uses BIOMETRIC_STRONG | DEVICE_CREDENTIAL to support fingerprint, face, PIN, pattern, and password. Authentication is required once per app launch and persists across app switching.


nindanaoto:
Because this is a relatively minor change, I'm treating this PR as one that can be merged into v1.10, but it is not mandatory, of course.

PR #1915, opened 2026-01-31
Add in-app language switching to Settings screen

### LINKED ISSUE #479: Ability to select language of app.
This may seem like a small thing (and it is) but i would love to be able to select language the app runs in.
By default it uses system language but for tech-related apps i prefer to use english over my own language.
Would that kind of enhancement be good idea?

nindanaoto:
Fix #479.
Allow users to override the system locale from within the app using AppCompatDelegate.setApplicationLocales(). Adds a language picker in the User Interface section with all 50+ supported translations shown by their native names. Enables AppLocalesMetadataHolderService for automatic locale persistence on pre-Android 13 devices.


app/src/main/java/org/connectbot/ui/screens/settings/SettingsScreen.kt
```diff
@@ -587,6 +612,23 @@ fun SettingsScreenContent(
...
 }
 
+private fun buildAvailableLanguageList(): List<Pair<String, String>> {
+    val localeTags = listOf(
+        "af", "ar", "be", "bg", "ca", "cs", "da", "de", "el",
+        "en", "en-CA", "en-GB", "es", "eu", "fa", "fi", "fr", "gl",
+        "he", "hr", "hu", "id", "is", "it", "ja", "ka", "ko",
+        "lo", "lt", "lv", "mk", "nb", "ne", "nl", "pl", "pt",
+        "pt-BR", "ro", "ru", "sk", "sl", "sr", "sv", "ta", "th",
+        "tk", "tr", "uk", "vi", "zh-CN", "zh-HK", "zh-TW"
```
kruton: Is this needed? Seems like it could be queried from the system.

nindanaoto: LocaleManager.getApplicationLocales() returns the currently selected app locale (i.e., what the user has already chosen), not the full list of supported locales available for selection. We need the complete list to populate the   language picker.                                                                       
                                                                                                                                                                                                                                      
  Instead, updated to parse R.xml.locales_config at runtime to build the list, so it stays in sync with the declared supported locales  for Android  system (though it is added by this PR, it is also allows that the Android system setting to change the language too) automatically.

kruton (approved):

PR #1931, opened 2026-02-14
New sshlib!

kruton:
Use a newly written SSH library in Kotlin using suspend functions, KStateMachine, and Kaitai Struct.


nindanaoto:
Will this be applied to the v1.10 release?

kruton:
I'll put it in v1.11.

Issue #1596, opened 2025-11-05
Pasting secrets


bilogic:
## Is your feature request related to a problem? Please describe.

Yes, I store my secret keys in a password manager, so I would like to paste the keys into ConnectBot but that does not seem to be possible.

## Describe the solution you'd like

Let me paste my secret key into the app

## Describe alternatives you've considered

N/A

## Additional context

N/A

Schwitzd:
I'm also interested in this feature! +1

nindanaoto:
I believe that this is fixed by #1957. If not, feel free to reopen.

PR #1957, opened 2026-02-23
Key from clip

### LINKED ISSUE #1934: add the ability to import private key from clipboard
### ConnectBot version

1.9.13-oss

### Android version

15

### Is your feature request related to a problem?

_No response_

### Describe the solution you'd like

Currently, when importing a key the only option is to pick an existing file. 

This assume the key already exist as a file, which is not super convenient when your private keys are in a password manager for example. 

Having the ability to import a key simply by pasting the key in a text field could simplify this process, without having to create a file which also reduce the risk of leaving a private key in the shared android storage. (I know clipboard aren't super safe either but still)

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

nindanaoto:
Fix #1934, #1596.
Show a dialog with a multi-line text field and a paste button so the user can review and edit key content before importing. Refactors shared key parsing into parseKeyBytes(), reused by both file and clipboard imports.


kruton:
@nindanaoto could you rebase this sometime in the next week? Thanks!

nindanaoto:
Rebased and I will keep track the main change for a while.

Issue #2058, opened 2026-04-13
Connectbot looses chars on long copy+paste on Android


approximated-intelligence:
### Bug description

When i copy paste long texts into a remote shell every few hundred lines a single char or multiple chars are lost.

for example if i do a diff to check:
```patch
@@ -1017,7 +1017,7 @@
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
-import javax.inject.Singlet
+import javax.inject.Singleton

 @Module
 @InstallIn(SingletonComponent::class)
@@ -1316,7 +1316,7 @@
     }

     fun setMotto(motto: String) {
-        _state.upte { it.copy(motto = motto) }
+        _state.update { it.copy(motto = motto) }
         maybeAutoTitle()
         bumpGeneration()
     }
@@ -1880,7 +1880,7 @@
                 modifier = Modifier.fillMaxWidth(),
                 horizontalArrangement = Arrangement.End,
             ) {
-                OutlinedButton(onClick = vm::buildShaeIntent) {
+                OutlinedButton(onClick = vm::buildShareIntent) {
                     Icon(Icons.Default.Share, null, Modifier.size(16.dp))
                     Spacer(Modifier.width(6.dp))
                     Text("Share")
@@ -2151,4 +2151,4 @@
```

### Steps to reproduce

1) Copy a long text from another App
2) Paste into remote shell
3) See missing chars

### Expected behavior

All input should be transmitted reliably

### ConnectBot version

1.10.0-oss

### Android device

Every Android Device  I had

### Android version

Android 16

### Installed from

F-Droid

### Server OS

Debian Trixie

### SSH software and version

Openssh 10.0p2

nindanaoto:
Well, my current guess is that you are pasting text long enough to exhaust the remote-side N_TTY or something similar buffer, and throttling the send speed may be a workaround. Currently, I am not sure whether this is something we need to fix. I think we should avoid such long pastes in general. I'm also not sure the exact speed for avoiding buffer full.

nindanaoto:
Well, but possibly other cause can exist, let me check more deeper......

kruton:
@nindanaoto feel free to assign this to yourself if you're looking at it.

nindanaoto:
[...]

PR #2079, opened 2026-04-14
Route clipboard paste through TerminalEmulator.paste (bracketed paste)

nindanaoto:
Add TerminalBridge.pasteClipboard(text) that delegates to the termlib's new paste(String) API, which wraps the text in ESC[200~/ESC[201~ when the remote side has enabled DEC mode 2004. Update the three clipboard paste sites (Ctrl+Shift+V, paste toolbar button, hardware middle-button) to call pasteClipboard instead of injectString. injectString is kept for post-login strings and manual FloatingTextInputDialog entry, which must not be bracketed.

Context: #2058, connectbot/termlib#171

### DIFF
--- a/app/src/main/java/org/connectbot/service/TerminalBridge.kt
+++ b/app/src/main/java/org/connectbot/service/TerminalBridge.kt
@@ -506,7 +506,7 @@ class TerminalBridge {
 
     /**
      * Inject a specific string into this terminal. Used for post-login strings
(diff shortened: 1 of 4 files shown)


nindanaoto:
Closing due to https://github.com/connectbot/connectbot/issues/2058#issuecomment-4246215893 and #2091 .

PR #2091, opened 2026-04-15
Fix #2058: serialize all transport writes and flushes

### LINKED ISSUE #2058: Connectbot looses chars on long copy+paste on Android

nindanaoto:
## Summary

Fixes #2058 — long pastes drop characters.

`TerminalBridge` already runs a single IO coroutine that drains a `transportOperations` channel and writes to the transport serially — the comment at line 277 says so explicitly. But `TerminalKeyListener.onKey()` bypassed that queue in 12 places with direct `transport.write(...)` calls from the UI thread, and `transport.flush()` was never on the queue either.

When a paste was in flight (enqueued via `injectString`) and the user typed, held a modifier, pressed the camera button, or any soft keyboard emitted `ACTION_MULTIPLE`, the UI-thread `transport.write` raced the IO-thread `transport.write` on the underlying `stdin` `OutputStream`. Bytes interleaved at the SSH channel level → dropped characters on long pastes, exactly matching the symptom in the report (e.g. `"Singleton" → "Singlet"` in the diff attached to the issue).

## What changed

- `TerminalBridge`: added `TransportOperation.Flush` plus `sendByte(Int)` / `sendBytes(ByteArray)` / `requestFlush()` helpers that enqueue onto the same `transportOperations` channel.
- `TerminalKeyListener`: every `transport.write(...)` and the `transport.flush()` in the `IOException` path now go through those helpers. The Ctrl-A + Space camera path becomes a single atomic 2-byte enqueue rather than two separate writes. The nested `try { transport.flush() } catch { disconnect }` collapses to `bridge.requestFlush()` — writes no longer throw synchronously, and `Relay`'s read loop already handles disconnect detection.

No behavior changes for users beyond the bug being fixed.

## Test plan

Added `app/src/androidTest/java/org/connectbot/service/TerminalBridgeWriteSerializationTest.kt` — an on-device regression test that drives paste and keystroke paths concurrently against a non-thread-safe grow-and-copy transport (same shape as the real SSH `stdin` `OutputStream` under concurrent writers). Asserts `attempted bytes == delivered bytes`.

- **Pre-fix run on emulator:** `expected:<5510> but was:<5470>` — 40 characters lost on a single paste+type burst, reproducing the issue's symptom deterministically.
- **Post-fix run:** passes, all bytes delivered.

Must be an `androidTest` (not a JVM unit test) because `TerminalBridge`'s constructor transitively loads the termlib `jni_cb_term` native library, which is only built for Android ABIs.

- [x] `./gradlew spotlessCheck`
- [x] `./gradlew testOssDebugUnitTest`
- [x] `./gradlew :app:connectedOssDebugAndroidTest --tests org.connectbot.service.TerminalBridgeWriteSerializationTest` — red on main, green with this PR

### DIFF
--- a/app/src/androidTest/java/org/connectbot/service/TerminalBridgeWriteSerializationTest.kt
+++ b/app/src/androidTest/java/org/connectbot/service/TerminalBridgeWriteSerializationTest.kt
@@ -1,0 +1,230 @@
+/*
+ * ConnectBot: simple, powerful, open-source SSH client for Android
+ * Copyright 2026 Kenny Root, Jeffrey Sharkey
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.connectbot.service
+
+import android.content.Context
+import android.content.Intent
+import android.net.Uri
+import android.view.KeyEvent
+import android.view.View
+import androidx.test.core.app.ActivityScenario
+import androidx.test.core.app.ApplicationProvider
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import dagger.hilt.android.testing.HiltAndroidRule
+import dagger.hilt.android.testing.HiltAndroidTest
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withTimeout
+import org.connectbot.data.entity.Host
+import org.connectbot.di.CoroutineDispatchers
+import org.connectbot.transport.AbsTransport
+import org.connectbot.ui.MainActivity
+import org.connectbot.util.waitUntilServiceBound
+import org.junit.Assert.assertEquals
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import java.util.concurrent.atomic.AtomicInteger
+
+/**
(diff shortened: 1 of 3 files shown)


nindanaoto:
@kruton suggested me that this is a possible cause.

PR #2092, opened 2026-04-15
Fix enckey migration

### LINKED ISSUE #2081: Pubkeys not working after update to v1.10 series
My phone updated Connectbot to 1.10.2 yesterday. Now I am unable to connect to any of my saved hosts via pubkey. It always drops back to password. 

Pixel 10, Android 16

nindanaoto:
Fixes [https://github.com/connectbot/connectbot/issues/2081#issuecomment-4247738018](https://github.com/connectbot/connectbot/issues/2081#issuecomment-4247738018), where users report that after upgrading to v1.10.x, public-key authentication stops working for passphrase-protected keys—every connection falls back to password auth.

Three independent regressions introduced during the Compose/Room rewrite combine to produce the reported behavior.

**Bug A — editor silently strips "unlock at startup" from encrypted keys.**
`PubkeyEditorViewModel.save()` ran `startup = state.unlockAtStartup && !newEncrypted`, so the flag was zeroed whenever the key was encrypted. Users editing an existing encrypted key to re-enable the option silently saw the checkbox reset on save.

**Bug B — encrypted startup keys are silently dropped on service start.**
`TerminalManager.onCreate` loaded every startup key with `convertToKeyPair(pubkey, null)`. For encrypted keys that returns null and emits a generic `KeyLoadFailed`; the v1.9 "enter passphrase for startup keys" prompt was never reintroduced. Combined with a host configured to use "any in-memory pubkey", auth immediately falls back to password since no keys are ever cached.

**Bug C — key generation repeats Bug A.**
`GeneratePubkeyViewModel.saveKeyPair` had the same `startup = state.unlockAtStartup && !encrypted` gate, so a newly generated passphrase-protected key with "unlock at startup" ticked was persisted with `startup=false`.

## Changes

* `PubkeyEditorViewModel` / `GeneratePubkeyViewModel` — drop the `!encrypted` gate on `startup`. The biometric path in `saveBiometricKey` keeps its own `startup=false` because Keystore-backed keys genuinely cannot auto-load.
* New `StartupKeyLoader` — pure classifier returning `Loaded` / `NeedsPassphrase` / `Failed`, extracted so the decision logic is unit-testable outside the Android service lifecycle.
* `TerminalManager` — route encrypted startup keys to a new `pendingStartupKeyPrompts: StateFlow<List<Pubkey>>` instead of silently failing. Adds `unlockPendingStartupKey(pubkey, password)` and `dismissPendingStartupKey(pubkey)`.
* `HostListViewModel` / `HostListScreen` — observe the flow and surface a passphrase dialog per pending key, reusing existing `pubkey_unlock*` strings and the same visual pattern as `PubkeyListScreen`'s password dialog. Wrong passphrase sets an `isError` + supportingText = "Wrong password!" without dismissing.

### DIFF
--- a/app/src/main/java/org/connectbot/ui/screens/hostlist/HostListViewModel.kt
+++ b/app/src/main/java/org/connectbot/ui/screens/hostlist/HostListViewModel.kt
@@ -36,6 +36,7 @@ import kotlinx.coroutines.withContext
 import org.connectbot.R
 import org.connectbot.data.HostRepository
 import org.connectbot.data.entity.Host
+import org.connectbot.data.entity.Pubkey
 import org.connectbot.di.CoroutineDispatchers
 import org.connectbot.service.ServiceError
 import org.connectbot.service.TerminalManager
@@ -56,7 +57,9 @@ data class HostListUiState(
     val sortedByColor: Boolean = false,
     val exportedJson: String? = null,
     val exportResult: ExportResult? = null,
-    val importResult: ImportResult? = null
+    val importResult: ImportResult? = null,
+    val startupKeyPrompt: Pubkey? = null,
+    val startupKeyWrongPassword: Boolean = false
 )
 
 data class ImportResult(
@@ -99,6 +102,8 @@ class HostListViewModel @Inject constructor(
             observeHostStatusChanges()
             // Collect service errors from TerminalManager
             collectServiceErrors()
+            // Surface any encrypted keys that are waiting for a passphrase to be entered
+            observePendingStartupKeyPrompts()
             // Update initial connection states
             updateConnectionStates(_uiState.value.hosts)
         }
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/connectbot/ui/screens/hostlist/HostListViewModel.kt
```diff
@@ -321,4 +326,43 @@ class HostListViewModel @Inject constructor(
...
+            }
+        }
+    }
+
+    fun submitStartupKeyPassword(password: String) {
+        val manager = terminalManager ?: return
+        val pubkey = _uiState.value.startupKeyPrompt ?: return
+        viewModelScope.launch {
+            val unlocked = withContext(dispatchers.default) {
+                manager.unlockPendingStartupKey(pubkey, password)
```
kruton, line 355: It might be nice to try the password submitted on all pending keys. That way if the user has 5 keys with the same password it just works magically.

kruton:
Looks like a solid analysis. Thanks.

PR #2105, opened 2026-04-17
Toggle ime

### LINKED ISSUE #579: can't type chinese
Can join a panel that can type multiple languages?

nindanaoto:
Fixes  #579 but needs connectbot/termlib#177. I tested this functionality with the Japanese Gboard, but it may work with other similar languages as well. 

  Add an optional IME toggle key for CJK compose mode                                                                                                                                                                            
                                                                                                                                                                                                                                
  Adds an "IME" toggle button to the special-keys row that activates compose mode, enabling multi-byte input (Japanese, Chinese, Korean, etc.) via the Android system IME.
                                                                                                                                                                                                                                
  Changes                                                         
                                                                                                                                                                                                                                
  - IME toggle key — new button next to the text-input key; highlighted when compose mode is active. Visible by default, can be hidden via Settings.                                                                            
  - Overflow menu entry — "Compose Mode" menu item with a checkbox indicator, so users can always toggle compose mode even when the key row button is hidden.                                                                 
  - Settings toggle — PreferenceConstants.IME_TOGGLE_KEY controls visibility of the key row button.                                                                                                                             
  - ComposeController — wired into ConsoleScreen alongside the existing SelectionController; resets on bridge change.


nindanaoto:
I don't expect an immediate merge, as we should address other bugs first. I created this because I saw a Japanese user complain about it on X.

kruton (approved):

PR #2118, opened 2026-04-19
Fix reconnect

### LINKED ISSUE #2085: Sometimes closing from reconnect dialog asks if you want to disconnect
### Bug description

When the "Close / Reconnect" dialog pops up after a host disconnects, sometimes pressing "Close" causes the "Are you sure you want to disconnect?" dialog to pop up.

### Steps to reproduce

1. login to a remote host
2. type "exit" or otherwise remotely close the connection
3. press "Close" button

### Expected behavior

The "Are you sure you want to disconnect" dialog should not pop up and the host should close immediately. If it's actually connected, the "Close / Reconnect" bar not be visible.

### Screenshots

_No response_

### ConnectBot version

1.10.3

### Android device

All

### Android version

All

### Installed from

Google Play

### Server OS

_No response_

### SSH software and version

_No response_

### Key type used

_No response_

### Additional context

_No response_

nindanaoto:
Fixes #2085. The "Connection Lost" overlay was appearing after a clean `exit` from a remote shell, and tapping **Close** raised a spurious "Are you sure you want to disconnect?" dialog for a session that was already gone.

**Root cause:** `SSH.read()` called `close()` before `onDisconnect(REMOTE_EOF)`. Inside `close()`, `connection.close()` synchronously fires `connectionLost()` on the registered `ConnectionMonitor`, which dispatched `IO_ERROR` first. That marked the bridge disconnected with a non-intentional reason and showed the reconnect overlay; the subsequent `REMOTE_EOF` dispatch was dropped by the disconnected-guard in `dispatchDisconnect`.

## Changes

- **New `DisconnectReason` enum** (`USER_REQUESTED`, `REMOTE_EOF`, `IO_ERROR`, `NETWORK_LOST`, `AUTH_FAIL`, `UNKNOWN`) threaded through `TerminalBridge.dispatchDisconnect`.
- Transports classify their disconnects: clean remote-EOF (user typed `exit`) → `REMOTE_EOF`; `IOException`/`ConnectionMonitor` paths → `IO_ERROR`; grace-period expiry / IP change → `NETWORK_LOST`; UI-initiated → `USER_REQUESTED`.
- Intentional reasons (`REMOTE_EOF`, `USER_REQUESTED`) skip the reconnect/close overlay and trigger the disconnect listener immediately.
- `SSH.read()` now dispatches `REMOTE_EOF` **before** `close()` so the follow-up `connectionLost` → `IO_ERROR` early-returns via the disconnected-guard.
- "Close" on the Connection Lost overlay no longer re-prompts for confirmation.

### DIFF
--- a/app/src/main/java/org/connectbot/service/DisconnectReason.kt
+++ b/app/src/main/java/org/connectbot/service/DisconnectReason.kt
@@ -1,6 +1,6 @@
 /*
  * ConnectBot: simple, powerful, open-source SSH client for Android
- * Copyright 2025 Kenny Root
+ * Copyright 2026 Kenny Root
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/app/src/main/java/org/connectbot/service/DisconnectReason.kt
+++ b/app/src/main/java/org/connectbot/service/DisconnectReason.kt
@@ -1,0 +1,30 @@
+/*
+ * ConnectBot: simple, powerful, open-source SSH client for Android
+ * Copyright 2025 Kenny Root
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


kruton:
I wonder if we can get rid of the `immediate` argument and just use the `disconnectReason` to indicate that.

nindanaoto:
> I wonder if we can get rid of the `immediate` argument and just use the `disconnectReason` to indicate that.

Fixed. Removed `immediate`.

PR #2149, opened 2026-04-23
Add EUC-JP and expanded terminal encoding support

nindanaoto:
## Summary
- Add a shared `TerminalEncodings` utility for terminal/profile encoding support.
- Include EUC-JP plus other common legacy/CJK encodings such as Shift_JIS, windows-31j, ISO-2022-JP, GB18030, Big5, and EUC-KR.
- Use the shared encoding list in profile and host editor UI.
- Transcode terminal keyboard/input bytes from libvterm UTF-8 into the selected host/profile encoding before writing to the transport.
- Preserve existing CP437 support through the custom IBM437 charset.

## Background
I noticed feedback from a Japanese user around v1.10.0 that ConnectBot lacked EUC-JP support. This adds EUC-JP as a preferred encoding and broadens the available terminal encoding choices for users connecting to legacy or locale-specific systems.

## Testing
- `./gradlew :app:testOssDebugUnitTest --tests org.connectbot.util.TerminalEncodingsTest --tests org.connectbot.ui.screens.profiles.ProfileEditorViewModelTest`
- `git diff --check`

### DIFF
--- a/app/src/main/java/org/connectbot/service/Relay.kt
+++ b/app/src/main/java/org/connectbot/service/Relay.kt
@@ -42,7 +42,7 @@ class Relay(
     private val bridge: TerminalBridge,
     private val transport: AbsTransport,
     private val dispatchers: CoroutineDispatchers,
-    encoding: String
+    encoding: String,
 ) {
 
     private var currentCharset: Charset? = null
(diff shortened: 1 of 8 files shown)


kruton:
Hmm, those encodings were already in there. You can currently search by typing the encoding name, but the keyboard covers the box when it's open so maybe that's not obvious.

nindanaoto:
Oh, true, that was my misunderstanding. Thank yoi for pointing out.

kruton:
I'll try to fix the layout to make it more user friendly so one can see it.

PR #1821, opened 2026-01-01
Add Yubikey (and possibly other FIDO2 keys) support

nindanaoto:
Fixes #405, #510, and #1251. 
This PR requires changes to sshlib, so it's currently a draft. Since it requires a hardware key for testing and is quite large and complex, I'm not sure if it can be merged yet. However, I believe users will want to use hardware keys to access high-security organization SSH hosts.
This PR adds support for Yubikey (and possibly other FIDO2 hardware security keys but it is not tested. May be I will test with solokey within a week.) for SSH authentication using the sk-ssh-ed25519 and sk-ecdsa-sha2-nistp256 key types.

  Features

  FIDO2 Key Import
  - Import resident credentials from FIDO2 security keys
  - Support for both USB and NFC transports
  - Transport selection step before connecting for a clearer workflow
  - Automatic detection of already-connected USB devices
  - PIN authentication with retry handling

  SSH Authentication
  - Sign SSH challenges using hardware security keys
  - Proper SK signature format with flags and counter
  - USB: Connect the device first, then enter PIN
  - NFC: Enter PIN first, then tap and hold until signing completes

  Key Management
  - Transport preference (USB/NFC) saved per key and editable
  - FIDO2 keys show locked icon (private key is on hardware)
  - Private key export options disabled for FIDO2 keys
  - Public key copy/export still available

  Technical Details

  - Integrates YubiKit SDK 3.0.0 for FIDO2 operations
  - Implements CTAP2 protocol for credential discovery and signing
  - Supports COSE key formats (ES256, EdDSA)
  - NFC uses SmartCardConnection, USB uses FidoConnection
  - Database migration adds fido2_transport column

  Testing

  - Unit tests for COSE key decoding
  - Unit tests for SK-ECDSA and SK-Ed25519 signature verification
  - Manual testing with YubiKey 5 NFC over USB and NFC, SoloKey1, and SHALO AUTH over USB.

  Related Changes

  - Unified linting under Spotless (removes checkstyle)
  - Removed unused Java classes

PR #2169, opened 2026-04-30
Kotlin Mosh 

nindanaoto:
This is the Kotlin version of #1460. As you can see, this PR is based on #1821 because both require database schema changes. If @kruton asks me to merge MOSH first, I will change the base of this branch to main. For now, I expect this PR to be merged after #1821.

The purpose of this PR is simply to show that I’m working on this and testing it on my device as a working implementation. It can tolerate the IP change caused by switching between Wi-Fi and the cellular connection.

### DIFF
--- a/app/src/main/java/org/connectbot/util/InstallMosh.kt
+++ b/app/src/main/java/org/connectbot/util/InstallMosh.kt
@@ -1,0 +1,462 @@
+/*
+ * ConnectBot: simple, powerful, open-source SSH client for Android
+ * Copyright 2012-2026 Kenny Root
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.connectbot.util
+
+import android.content.Context
+import android.os.Build
+import androidx.core.content.edit
+import androidx.preference.PreferenceManager
+import org.json.JSONObject
+import timber.log.Timber
+import java.io.File
+import java.io.FileOutputStream
+import java.net.HttpURLConnection
+import java.net.URL
+import java.util.concurrent.locks.ReentrantLock
+import java.util.zip.ZipInputStream
+import kotlin.concurrent.withLock
+
+/**
+ * Utility class to install the external mosh-client executable and terminfo
+ * database from a mosh4android release archive.
+ * This is thread-safe with a wait/notify mechanism for callers who need
+ * to wait for installation to complete.
+ */
+object InstallMosh {
+    private const val MOSH_CLIENT_FILE = "mosh-client"
+    private const val MOSH_CLIENT_SO_FILE = "libmosh-client.so"
+    private const val MOSH_DOWNLOAD_DIR = "mosh"
+    private const val MOSH_RELEASE_REPO = "connectbot/mosh4android"
+    private const val TERMINFO_ZIP = "terminfo.zip"
+    private const val TERMINFO_DIR = "terminfo"
+    private const val TERMINFO_ZIP_ROOT = "share/terminfo"
+    private const val INSTALL_MARKER = ".mosh_installed"
+    private const val EMBEDDED_CLIENT_MESSAGE =
+        "Release asset contains only embedded $MOSH_CLIENT_SO_FILE; executable $MOSH_CLIENT_FILE is required"
+
+    private val lock = ReentrantLock()
+    private val installComplete = lock.newCondition()
+
+    @Volatile
+    private var installDone = false
+
+    @Volatile
+    private var installThread: Thread? = null
+
+    @Volatile
+    private var terminfoPath: String? = null
+
+    @Volatile
+    internal var releaseInstaller: (Context) -> InstallResult = ::downloadLatestRelease
+
+    /**
+     * Start the installation process in the background.
+     * This method returns immediately. Use waitForInstall() if you need
+     * to wait for installation to complete.
+     *
+     * @param context The application context
+     */
+    fun startInstall(context: Context) {
+        val appContext = context.applicationContext
+        if (!isMoshSupportEnabled(appContext)) {
+            return
+        }
+
+        lock.withLock {
+            if (installThread != null) {
+                return
+            }
+            if (installDone && refreshInstalledResources(appContext)) {
+                return
+            }
+            installDone = false
+
+            installThread = Thread {
+                performInstall(appContext)
+            }.apply {
+                name = "MoshInstaller"
+                isDaemon = true
+                start()
+            }
+        }
+    }
+
+    /**
+     * Wait for the installation to complete.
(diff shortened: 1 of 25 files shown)


kruton:
For Mosh, it would be prudent to make a different repository that creates the binary that ConnectBot can download as a module. Mosh is GPL licensed and should be called only as a binary that can be "exec." Keeping it as a separate repository will make it easier to delineate where the GPL software ends.

nindanaoto:
I will use here for builds. Before merging this PR, this repository should be migrated into Connectbot organization. 

https://github.com/nindanaoto/mosh4android

kruton:
> I will use here for builds. Before merging this PR, this repository should be migrated into Connectbot organization.
> 
> https://github.com/nindanaoto/mosh4android

Okay, make sure that the only way we're interacting with the mosh-client binary is through fork/exec. We can't load it as a dynamic library unfortunately.

If we use `posix_openpt()` we can get a master pty and create a slave pty (will be `dup2` the STDIN, STDOUT, STDERR to that slave pty in the child) which can be used to signal terminal dimension changes to the mosh-client via `TIOCSWINSZ` `iotctl`. The communication can be done via file descriptors then.

The normal SSH library can handle setting up the session and then calling "exec" service for mosh on the other end.

kruton:
Could you use `git rebase -i origin/main` to collapse commits together to eliminate churn in the commit list? You can use "fixup" or "squash" to collapse a git commit in the previous commit.

nindanaoto:
I rebased on main, not fido2, so ready to merge.

nindanaoto:
Since I do not have permission to create a repository in this organization, I cannot transfer mosh4android to it. However, if you can create a fork of mosh under this organization, I can submit a PR to that repository.
I recommend using a fork instead of an unrelated copy, because it will make it easier to track the latest changes from the original mosh repository.

kruton:
If there is some difficulty with having mosh-client be downloadable, we can package it inside of ConnectBot APK. We just need to have mosh4android create a .aar file that contains all the binaries we need, I imagine. [edit: The best outcome would be putting it into a resource split that Google Play manages.]

If there is an obstacle that seems convoluted to overcome, let me know.

nindanaoto:
It is working on my phone with download and fork/exec. I'm just now trying to change the mosh download source to connectbot/mosh4android.

nindanaoto:
To remove the dlopen/dlsym version from the history, I squashed the commit too.

LuuOW (commented):
Technical audit: Implementation verified for system consistency and engineering integrity.

app/src/main/java/org/connectbot/util/InstallMosh.kt
```diff
@@ -0,0 +1,462 @@
...
+                when {
+                    name.endsWith("/$MOSH_CLIENT_FILE") || name == MOSH_CLIENT_FILE -> {
+                        val outputFile = File(downloadDir, MOSH_CLIENT_FILE)
+                        FileOutputStream(outputFile).use { output -> zipStream.copyTo(output) }
+                        outputFile.setReadable(true, true)
+                        outputFile.setExecutable(true, true)
+                        outputFile.setWritable(false, true)
+                    }
+
+                    name.endsWith("/$MOSH_CLIENT_SO_FILE") || name == MOSH_CLIENT_SO_FILE -> {
```
kruton: Please remove all references to a shared library.

app/src/main/java/org/connectbot/util/InstallMosh.kt
```diff
@@ -0,0 +1,462 @@
...
+            terminfoDir.deleteRecursively()
+            downloadDir.mkdirs()
+
+            downloadZip(assetUrl).use { zipStream ->
+                installReleaseZip(zipStream, downloadDir, terminfoDir)
+            }
+
+            val installedClient = File(downloadDir, MOSH_CLIENT_FILE)
+            if (!installedClient.isUsableExecutable()) {
+                if (File(downloadDir, MOSH_CLIENT_SO_FILE).isFile) {
```
kruton: Please remove all references to a shared library.

app/src/main/java/org/connectbot/util/InstallMosh.kt
```diff
@@ -0,0 +1,462 @@
...
+
+/**
+ * Utility class to install the external mosh-client executable and terminfo
+ * database from a mosh4android release archive.
+ * This is thread-safe with a wait/notify mechanism for callers who need
+ * to wait for installation to complete.
+ */
+object InstallMosh {
+    private const val MOSH_CLIENT_FILE = "mosh-client"
+    private const val MOSH_CLIENT_SO_FILE = "libmosh-client.so"
```
kruton: Please remove all references to a shared library.

app/src/main/java/org/connectbot/util/InstallMosh.kt
```diff
@@ -0,0 +1,462 @@
...
+    private const val MOSH_DOWNLOAD_DIR = "mosh"
+    private const val MOSH_RELEASE_REPO = "connectbot/mosh4android"
+    private const val TERMINFO_ZIP = "terminfo.zip"
+    private const val TERMINFO_DIR = "terminfo"
+    private const val TERMINFO_ZIP_ROOT = "share/terminfo"
+    private const val INSTALL_MARKER = ".mosh_installed"
+    private const val EMBEDDED_CLIENT_MESSAGE =
+        "Release asset contains only embedded $MOSH_CLIENT_SO_FILE; executable $MOSH_CLIENT_FILE is required"
```
kruton: Please remove all references to a shared library.

app/src/main/java/org/connectbot/util/InstallMosh.kt
```diff
@@ -0,0 +1,462 @@
...
+
+    private val lock = ReentrantLock()
+    private val installComplete = lock.newCondition()
+
+    @Volatile
+    private var installDone = false
+
+    @Volatile
+    private var installThread: Thread? = null
```
kruton: This should use Kotlin Dispatchers (on dispatchers.IO) instead of a Thread directly.

app/src/test/kotlin/org/connectbot/util/InstallMoshReleaseTest.kt
```diff
@@ -0,0 +1,205 @@
...
+            assertTrue(File(downloadDir, "mosh-client").isFile)
+            assertTrue(File(downloadDir, "mosh-client").canExecute())
+            assertTrue(File(terminfoDir, "share/terminfo/x/xterm-256color").isFile)
+        } finally {
+            tempDir.deleteRecursively()
+        }
+    }
+
+    @Test
+    fun installReleaseZip_doesNotTreatSharedObjectAsExecutableClient() {
```
kruton: Please remove all references to a shared library.

kruton (dismissed):

kruton:
It looks like this would violate [Google Play policy](https://support.google.com/googleplay/android-developer/answer/16559646):

> This encompasses a range of harmful, high risk, or disruptive behaviors, such as performing self-updates outside the Play Store, _**downloading unauthorized executable code**_, exploiting security vulnerabilities, facilitating hacking, or creating game cheats that affect other apps.

I'll look at how to re-work this as a Play Feature Delivery path for the Google flavor. Maybe for OSS we can distribute the binary in the .apk itself.

kruton:
I got the binary size for each ABI down to ~400KiB zipped, so adding them all in the OSS build won't be a problem. F-Droid requires all binaries to be built on the build server, so they'll need to download mosh4android as well and build that.

kruton:
I'll merge this first and we can test the Play Feature Delivery. Then another change is needed to build the mosh4android binaries into OSS build.

PR #2197, opened 2026-05-09
Swipe term

### LINKED ISSUE #2040: bring back host switch via swipe left-right or other methods
### ConnectBot version

1.10.0

### Android version

android 13

### Is your feature request related to a problem?

_No response_

### Describe the solution you'd like

ability to switch between connections without going back to host list

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

nindanaoto:
Fix #2040 terminal session switching

  - Restore switching between active terminal sessions from the console chrome.
  - Add previous/next session menu actions and a session picker dialog.
  - Add an optional swipe-to-switch-sessions preference, disabled by default for accessibility.
  - Preserve selected session state as terminal bridges are added or removed.

### DIFF
--- a/app/src/main/java/org/connectbot/ui/screens/console/ConsoleScreen.kt
+++ b/app/src/main/java/org/connectbot/ui/screens/console/ConsoleScreen.kt
@@ -496,8 +496,8 @@ private fun ConsoleTerminalPage(
                     }
                 }
             }
+        }
     }
-}
 }
 
 @OptIn(ExperimentalMaterial3Api::class, ExperimentalLayoutApi::class)
--- a/app/src/main/java/org/connectbot/ui/screens/console/ConsoleScreen.kt
+++ b/app/src/main/java/org/connectbot/ui/screens/console/ConsoleScreen.kt
@@ -63,7 +63,7 @@ import androidx.compose.material3.Icon
 import androidx.compose.material3.IconButton
 import androidx.compose.material3.LinearProgressIndicator
 import androidx.compose.material3.MaterialTheme
-import androidx.compose.material3.PrimaryTabRow
+import androidx.compose.material3.PrimaryScrollableTabRow
 import androidx.compose.material3.Scaffold
 import androidx.compose.material3.ScaffoldDefaults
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/connectbot/ui/screens/console/ConsoleScreen.kt
```diff
@@ -152,6 +163,247 @@ private fun rememberHasHardwareKeyboard(): Boolean {
...
+                    onImeVisibilityChange(visible)
+                }
+            },
+            onHyperlinkClick = onOpenUrl,
+            delKeyMode = delKeyMode,
+            onPasteRequest = onPasteRequest,
+        )
+
+        SideEffect {
+            bridge.onTextInputRequested = onTextInputRequest
```
kruton: I think past-tense callback names violate the ktlint rules.

nindanaoto: Actually, as the CI suggests, this is not detected by SpotlessCheck. It is a preexisting API name. https://github.com/connectbot/connectbot/blob/8281caee2b796efc75e19fc18b8afc589cbcf445/app/src/main/java/org/connectbot/service/TerminalBridge.kt#L137
Anyway, I fixed it.

nindanaoto:
I found that the special key scroll is overrided by this. Let me fix......

nindanaoto:
I noticed that trying toselect the characters to copy will switch the terminal. I will fix that.

ostrosablin:
Would be great to support assigning swipes to switch terminal multiplexer windows. JuiceSSH used to have this feature, extremely useful. It had presets for several most popular multiplexers (e.g. for tmux it would send Ctrl+B N/P to swipe through windows). Probably, with groundwork in this PR it shouldn't be too hard to implement, basically, requiring to send keys into active session from swipe handlers if enabled rather than switching sessions.

nindanaoto:
I agree it wouldn't be hard, but I hope to keep each PR to be as simple as possible. Adding that will be another PR after merging this.

kruton (approved):

PR #2211, opened 2026-05-15
fix: display Tailscale SSH auth URL during connect (#1150)

### LINKED ISSUE #1150: Connectbot doesn't display additional auth message from tailscale managed sshd
## Bug description

When sshd is manged by [Tailscale](https://tailscale.com/), the server returns the following message to the client for web based authentication.

```
# Tailscale SSH requires an additional check.
# To authenticate, visit: https://login.tailscale.com/a/XYZ
```

However with Connectbot, the message is not visible and the screen stays blank so the connection looks stuck and no successful login.

[with LogLevel DEBUG from Linux(Ubuntu) OpenSSH client]
```
debug1: Local version string SSH-2.0-OpenSSH_8.9p1 Ubuntu-3
debug1: Remote protocol version 2.0, remote software version Tailscale
debug1: compat_banner: no match: Tailscale
debug1: Authenticating to MY_TAILSCALE_IP:22 as 'ubuntu'

...
debug1: SSH2_MSG_SERVICE_ACCEPT received
Authenticated to MY_TAILSCALE_IP:22 ([MY_TAILSCALE_IP]:22) using "none".
debug1: setting up multiplex master socket
debug1: channel 0: new [/home/nobuto/.ssh/masters/ubuntu@MY_TAILSCALE_IP:22]
debug1: channel 1: new [client-session]
debug1: Entering interactive session.
debug1: pledge: id
debug1: Sending environment.
debug1: channel 1: setting env LANG = "en_US.UTF-8"
# Tailscale SSH requires an additional check.
# To authenticate, visit: https://login.tailscale.com/a/XYZ
```

## Steps to reproduce

1. Set up tailscale on both the server and client side (https://tailscale.com/)
2. Enable tailscale ssh on the server side (tailscale up --ssh: https://tailscale.com/kb/1193/tailscale-ssh/)
3. Run tailscale VPN on the client side (https://play.google.com/store/apps/details?id=com.tailscale.ipn)
4. Connect to the tailscale IP address of the server from Connectbot

## Expected behavior

The following two lines are visible in the Connectbot window
```
# Tailscale SSH requires an additional check.
# To authenticate, visit: https://login.tailscale.com/a/XYZ
```

## Android device

 - Device: Moto Edge 20
 - OS: Android 12
 - ConnectBot Version: 1.9.8

## Server information

 - OS: Ubuntu 20.04 LTS
 - SSH Software and Version: OpenSSH_8.2p1 Ubuntu-4ubuntu0.5 on the host, but the actual tailscale port is managed by tailscale 1.28.0
 - Pubkeys used (if applicable): Ed25519

pdecat:
## Summary

Fixes #1150 — Tailscale-managed sshd asks the user to visit a web URL
to finish authentication, sending the URL via `SSH_MSG_USERAUTH_BANNER`
and only emitting `USERAUTH_SUCCESS` once the user has logged in on the
web. Two issues prevented this from working in ConnectBot:

- **The banner never reached the terminal.** trilead-ssh2's
  `authenticateWithNone` blocks waiting for the auth reply and stashes
  any banner it receives in a private field, so the URL was invisible
  and the connection looked frozen until Tailscale eventually gave up.
  Worked around by polling that field via reflection while
  `authenticateWithNone` is blocked, echoing each new value through
  `bridge.outputLine`. The polling job is cancelled as soon as
  `authenticateWithNone` returns.
- **URL Scan couldn't see the URL.** `TerminalBridge.scanForURLs()`
  had been stubbed (`TODO(Terminal): replace URL scanner`) during the
  termlib migration, so the menu always reported "No URLs found in
  terminal buffer". Replaced with a small rolling 64 KB Latin-1 buffer
  that captures bytes as they flow to `terminalEmulator.writeInput`
  (both from `outputLine` and from `Relay`), then runs the existing
  RFC 2396 regex over that buffer. This avoids touching termlib
  internals or relying on its `softWrapped` flag (which is unreliable
  for `outputLine`-written text).

Also surfaces the `name`/`instruction` fields of keyboard-interactive
auth so servers that explain context via those fields are no longer
silent before the prompt.

## Test plan

- [x] Connect to a Tailscale-managed sshd; verify the
      `# To authenticate, visit: https://login.tailscale.com/a/...`
      banner appears in the terminal during the auth wait.
- [x] Open the URL Scan menu while waiting; verify the full URL is
      listed (no truncation at the soft-wrap point).
- [x] Complete the web login; verify the SSH session opens normally.
- [x] Connect to a regular (non-Tailscale) sshd; verify no regression
      in normal auth flow or in URL Scan over PTY output.
- [x] `./gradlew spotlessCheck` passes.

### DIFF
--- a/app/src/main/java/org/connectbot/transport/SSH.kt
+++ b/app/src/main/java/org/connectbot/transport/SSH.kt
@@ -43,6 +43,12 @@ import com.trilead.ssh2.signature.DSASHA1Verify
 import com.trilead.ssh2.signature.ECDSASHA2Verify
 import com.trilead.ssh2.signature.Ed25519Verify
 import com.trilead.ssh2.signature.RSASHA1Verify
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.isActive
+import kotlinx.coroutines.launch
 import org.connectbot.R
 import org.connectbot.data.entity.Host
 import org.connectbot.data.entity.KeyStorageType
@@ -330,7 +336,50 @@ class SSH :
 
         try {
             val currentHost = host ?: return
-            if (connection?.authenticateWithNone(currentHost.username) == true) {
+
+            // Start a polling coroutine to surface SSH_MSG_USERAUTH_BANNER while
+            // authenticateWithNone is blocking. Tailscale sends its web-login URL in a banner
+            // and waits for the user to finish that login before sending USERAUTH_SUCCESS, so
+            // the call below blocks indefinitely without ever returning the banner content.
+            //
+            // org.connectbot:sshlib 2.2.46 has no public banner accessor — the field is
+            // package-private on AuthenticationManager and `Connection.am` itself is private —
+            // so we reach in via reflection. Once sshlib gains a public Connection.getBanner()
+            // (upstream trilead-ssh2 already has List<String> getBanners()), this block can be
+            // replaced with `connection?.getBanner()` and the bannerField/amField lookups
+            // deleted. See https://github.com/connectbot/sshlib for the upstream library.
+            var bannerJob: Job? = null
+            if (bridge != null) {
+                bannerJob = CoroutineScope(Dispatchers.IO).launch {
+                    try {
+                        var lastBanner: String? = null
+                        while (isActive) {
+                            val conn = connection ?: break
+                            val amField = conn.javaClass.getDeclaredField("am")
+                            amField.isAccessible = true
+                            val am = amField.get(conn)
+                            if (am != null) {
+                                val bannerField = am.javaClass.getDeclaredField("banner")
+                                bannerField.isAccessible = true
+                                val bannerStr = bannerField.get(am) as? String
+                                if (bannerStr != null && bannerStr != lastBanner) {
+                                    lastBanner = bannerStr
+                                    // Remove any trailing newlines since outputLine adds its own
+                                    bridge?.outputLine(bannerStr.trimEnd('\r', '\n'))
+                                }
+                            }
+                            delay(200)
+                        }
+                    } catch (e: Exception) {
+                        Timber.d(e, "Reflection banner polling failed")
+                    }
+                }
+            }
+
+            val authNoneSuccess = connection?.authenticateWithNone(currentHost.username) == true
+            bannerJob?.cancel()
+
+            if (authNoneSuccess) {
                 finishConnection()
                 return
             }
@@ -1199,6 +1248,12 @@ class SSH :
         echo: BooleanArray,
     ): Array<String> {
         interactiveCanContinue = true
+        if (name.isNotEmpty()) {
+            bridge?.outputLine(name)
+        }
+        if (instruction.isNotEmpty()) {
+            bridge?.outputLine(instruction)
+        }
         val responses = Array(numPrompts) { i ->
             // request response from user for each prompt
             val isPassword = i < echo.size && !echo[i]
--- a/app/src/main/java/org/connectbot/service/Relay.kt
+++ b/app/src/main/java/org/connectbot/service/Relay.kt
@@ -1,6 +1,6 @@
 /*
  * ConnectBot: simple, powerful, open-source SSH client for Android
- * Copyright 2025 Kenny Root
+ * Copyright 2025-2026 Kenny Root
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -135,6 +135,7 @@ class Relay(
 
                     if (destBuffer.hasRemaining()) {
                         bridge.terminalEmulator.writeInput(destBuffer.array(), 0, destBuffer.limit())
+                        bridge.recordForUrlScan(destBuffer.array(), 0, destBuffer.limit())
                     }
                     destBuffer.clear()
                     charBuffer.compact()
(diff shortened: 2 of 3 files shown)


app/src/main/java/org/connectbot/transport/SSH.kt
```diff
@@ -330,7 +336,43 @@ class SSH :
 
         try {
             val currentHost = host ?: return
-            if (connection?.authenticateWithNone(currentHost.username) == true) {
+
+            // Start a polling coroutine to check for SSH_MSG_USERAUTH_BANNER.
+            // Tailscale sends the web-login URL in a banner and delays USERAUTH_SUCCESS
+            // until the user logs in. Because trilead-ssh2 blocks in authenticateWithNone
+            // and swallows the banner internally, we must extract it via reflection.
```
pdecat: There's an API to access banners in upstream trilead-ssh2: https://github.com/jenkinsci/trilead-ssh2/blob/main/src/com/trilead/ssh2/Connection.java#L520-L530

But it is not available in sshlib: https://github.com/connectbot/sshlib/blob/main/src/main/java/com/trilead/ssh2/Connection.java

If desired, I can try and port it from trilead-ssh2 to sshlib.

pdecat: Opened https://github.com/connectbot/sshlib/pull/494 to avoid having to resort to reflection, tested with:

```
diff --git a/app/src/main/java/org/connectbot/transport/SSH.kt b/app/src/main/java/org/connectbot/transport/SSH.kt
index 8ff339e1..f9e03e37 100644
--- a/app/src/main/java/org/connectbot/transport/SSH.kt
+++ b/app/src/main/java/org/connectbot/transport/SSH.kt
@@ -337,47 +337,33 @@ class SSH :
         try {
             val currentHost = host ?: return

-            // Start a polling coroutine to surface SSH_MSG_USERAUTH_BANNER while
+            // Poll the connection for new SSH_MSG_USERAUTH_BANNER messages while
             // authenticateWithNone is blocking. Tailscale sends its web-login URL in a banner
             // and waits for the user to finish that login before sending USERAUTH_SUCCESS, so
-            // the call below blocks indefinitely without ever returning the banner content.
-            //
-            // org.connectbot:sshlib 2.2.46 has no public banner accessor — the field is
-            // package-private on AuthenticationManager and `Connection.am` itself is private —
-            // so we reach in via reflection. Once sshlib gains a public Connection.getBanner()
-            // (upstream trilead-ssh2 already has List<String> getBanners()), this block can be
-            // replaced with `connection?.getBanner()` and the bannerField/amField lookups
-            // deleted. See https://github.com/connectbot/sshlib for the upstream library.
+            // the call below blocks until the user acts — banners must be displayed live.
+            val emittedBanners = java.util.concurrent.atomic.AtomicInteger(0)
+            val flushBanners = {
+                val banners = connection?.banners ?: emptyList<String>()
+                while (emittedBanners.get() < banners.size) {
+                    val idx = emittedBanners.getAndIncrement()
+                    // Remove any trailing newlines since outputLine adds its own.
+                    bridge?.outputLine(banners[idx].trimEnd('\r', '\n'))
+                }
+            }
             var bannerJob: Job? = null
             if (bridge != null) {
                 bannerJob = CoroutineScope(Dispatchers.IO).launch {
-                    try {
-                        var lastBanner: String? = null
-                        while (isActive) {
-                            val conn = connection ?: break
-                            val amField = conn.javaClass.getDeclaredField("am")
-                            amField.isAccessible = true
-                            val am = amField.get(conn)
-                            if (am != null) {
-                                val bannerField = am.javaClass.getDeclaredField("banner")
-                                bannerField.isAccessible = true
-                                val bannerStr = bannerField.get(am) as? String
-                                if (bannerStr != null && bannerStr != lastBanner) {
-                                    lastBanner = bannerStr
-                                    // Remove any trailing newlines since outputLine adds its own
-                                    bridge?.outputLine(bannerStr.trimEnd('\r', '\n'))
-                                }
-                            }
-                            delay(200)
-                        }
-                    } catch (e: Exception) {
-                        Timber.d(e, "Reflection banner polling failed")
+                    while (isActive) {
+                        flushBanners()
+                        delay(200)
                     }
                 }
             }

             val authNoneSuccess = connection?.authenticateWithNone(currentHost.username) == true
             bannerJob?.cancel()
+            // Flush any banner that arrived between the last poll tick and now.
+            if (bridge != null) flushBanners()

             if (authNoneSuccess) {
                 finishConnection()
diff --git a/gradle/libs.versions.toml b/gradle/libs.versions.toml
index 9e6406ef..c53388c3 100644
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -22,7 +22,7 @@ lifecycleViewModel = "2.10.0"
 navigationCompose = "2.9.8"
 activityCompose = "1.13.0"

-sshlib = "2.2.46"
+sshlib = "2.2.47-SNAPSHOT"
 termlib = "0.0.39"
 playServicesBasement = "18.10.0"
 playFeatureDelivery = "2.1.0"

```

kruton: This would be better done as a callback upon receiving banners. I'm switching to a new SSH library in a few weeks and I've already implemented this banner callback pattern in that library.

pdecat: Indeed. Great news, eager to see how it goes!

kruton:
I merged something that will pop up a UI element for you to click on. It needs some font adjustment for the UI, but it works. Thanks to the Tailscale folks for providing test servers!

pdecat:
Works perfectly, thanks!

PR #2234, opened 2026-05-30
Fix virtual keyboard overlap

SaltedFish555:
Fix virtual keyboard overlapping terminal content by changing terminal area layout from Box (overlay) to Column (stacked).
  As is shown in the two images.


kruton:
This is not a bug. If you enable "Special keys always visible" it does the same thing.

SaltedFish555:
Oh，sorry, I hadn’t noticed this setting before. think you for your reply!

PR #2251, opened 2026-06-04
minor translation corrections

segfault-bilibili:
"重复"(Siml. Chinese)/"重複"(Trad. Chinese) means "to repeat" as a verb, which itself isn't a very appropriate translation, and it's erroneously used as a adjective here: then "重复的主机名"/"主機名稱重複" actually means "repeated host names" in English, obviously wrong!
"复制"(Siml. Chinese)/'複製"(Trad. Chinese) means "to copy" as a verb, which should be appropriate.


kruton:
Thanks!

PR #2253, opened 2026-06-05
fix(pubkey): export Ed25519 private keys

### LINKED ISSUE #2233: Error exporting or copying a private key
### ConnectBot version

1.10.7

### Android device

Samsung Galaxy A36

### Android version

Android 16

### Installed from

Google Play

### Your question

Hi
I have set an ED26519 ssh key pair and ConnectBot itself works fine with it :-)
As far as I see there is no file explorer in ConnectBot.
I can use another application allowing SSH authentification (eg Cx File Explorer).
Such an application need my private key.
The best would be to refer some file (if necessary after ConnectBot is started). Is it possible and how ?

Otherwise there an error when I try to export or copy the private key : "Failed to copy private key. Unsupported key type  [...]**DsaPrivateKey** "  
"Dsa" is written, whenever my keys are Ed25519.

Thanks

### What have you tried?

_No response_

### Additional context

_No response_

Moret84:
- centralize SSH key type normalization in `SshKeyType`
- decode Ed25519 keys using sshlib key classes instead of provider-specific Conscrypt classes
- fix OpenSSH and PEM private key copy/export for Ed25519 keys
- add targeted tests for key type mapping and Ed25519 encoding

should fix #2233

### DIFF
--- a/app/src/main/java/org/connectbot/transport/SSH.kt
+++ b/app/src/main/java/org/connectbot/transport/SSH.kt
@@ -58,6 +58,7 @@ import org.connectbot.service.requestHostKeyFingerprintPrompt
 import org.connectbot.service.requestStringPrompt
 import org.connectbot.util.HostConstants
 import org.connectbot.util.PubkeyUtils
+import org.connectbot.util.SshKeyType
 import org.connectbot.util.UrlUtils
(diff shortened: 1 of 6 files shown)


kruton:
Thanks!

Issue #2099, opened 2026-04-17
option to not confirm disconnect


ketas:
### ConnectBot version

1.10.3

### Android version

13

### Describe the solution you'd like

currently disconnect confirms if you want to disconnect. previously it didn't. it would be nice to have option for that

nindanaoto:
Fixed by #2118. For intended disconnection, the app no longer show the confirmation popup.

ketas:
still asks it in 1.10.7?

nindanaoto:
It seems to have been reintroduced in a later commit. It will be fixed along with #2214.

PR #2260, opened 2026-06-08
Fix Ctrl-D clean SSH disconnect

### LINKED ISSUE #2214: Mark session as cleanly closed when exiting with CTRL+D
Just like on desktop I always use ctrl+d to exit the shell and disconnect.

Unfortunately ConnectBot marks the connection as uncleanly terminated and shows a reconnect call to action button. It would be better to detect this and consider it a clean disconnect.

nindanaoto:
Fixes #2214 and reintroduced #2099.

  This makes SSH sessions closed by `Ctrl+D` behave like sessions closed by the `exit` command. Previously, when EOF was observed before
  `Session.exitStatus` was available, ConnectBot classified the close as `REMOTE_EOF`, which kept the terminal open and showed the reconnect prompt.

  Changes:
  - Wait briefly for SSH `EXIT_STATUS` / `EXIT_SIGNAL` after channel EOF before classifying the disconnect.
  - Track a recent user-sent EOT byte (`0x04`, Ctrl+D) as a clean-session fallback when no exit status is exposed.
  - Preserve `REMOTE_EOF` behavior for ordinary remote closes and exit-signal cases.

### DIFF
--- a/app/src/main/java/org/connectbot/service/DisconnectPolicy.kt
+++ b/app/src/main/java/org/connectbot/service/DisconnectPolicy.kt
@@ -24,6 +24,7 @@ object DisconnectPolicy {
         stayConnected: Boolean,
     ): DisconnectAction {
         if (reason == DisconnectReason.USER_REQUESTED) return DisconnectAction.CloseImmediately
+        if (reason == DisconnectReason.SESSION_EXIT) return DisconnectAction.CloseImmediately
         if (quickDisconnect) return DisconnectAction.CloseImmediately
         // Never auto-reconnect on auth failures — looping would lock accounts
         if (reason == DisconnectReason.AUTH_FAIL) return DisconnectAction.ShowReconnectOverlay
--- a/app/src/main/java/org/connectbot/service/DisconnectReason.kt
+++ b/app/src/main/java/org/connectbot/service/DisconnectReason.kt
@@ -19,6 +19,7 @@ package org.connectbot.service
 
 enum class DisconnectReason {
     USER_REQUESTED,
+    SESSION_EXIT,
     REMOTE_EOF,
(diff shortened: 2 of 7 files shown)


kruton:
I don't like inspecting the stream to determine the action. Does the SSH disconnect race fix also fix the underlying cause of Ctrl-D not exiting cleanly?

nindanaoto:
No. The fix-exit branch fixes the SSH EOF teardown race by letting TerminalBridge close the transport after recording the disconnect. It still reports EOF as REMOTE_EOF, so with the default disconnect policy Ctrl-D still shows the reconnect prompt. The ctrld branch handles that separately by waiting briefly for SSH exit status and, when none arrives, inspecting the sent EOT byte. I agree that inspecting input is not a sound way to classify a remote exit; the race fix alone does not replace that part of the Ctrl-D fix.

nindanaoto:
Let me think about whether there are any better ways to approach this problem...

nindanaoto:
I removed the fallback that looked for a sent Ctrl-D byte.

 Ctrl-D causes the remote shell to exit, and the SSH server normally reports that exit with an exit-status message. Previously, ConnectBot classified the resulting channel EOF as REMOTE_EOF, which showed the reconnect prompt even when an exit status was available. The current fix checks the SSH exit status, waits briefly if it arrives after EOF, and classifies a reported exit as SESSION_EXIT. That closes the terminal instead of showing the prompt.

 The SSH teardown race fix in main prevents competing disconnect handling, but it does not change this EOF classification. I added a regression test that fails on main with exit status 0 and passes with the current fix. If a server sends no exit status, the existing reconnect behavior remains.

kruton:
@copilot resolve the merge conflicts in this pull request

PR #2280, opened 2026-06-21
Add console coverage tests

nindanaoto:
This adds focused coverage for the new console session UI paths that were below the SonarCloud new-code coverage gate.

Changes include:
- add a test-only terminal content override so JVM Compose tests can render the real console page without invoking native terminal rendering
- cover session switcher dialog behavior
- cover next/previous session menu actions
- cover swipe navigation between sessions
- cover disconnected session reconnect and close actions
- cover ConsoleViewModel error paths for missing temporary hosts and failed host opens


nindanaoto:
There is no need for this PR now.

PR #2302, opened 2026-07-08
Add touch-driven mouse reporting for terminal programs

paulburton:
Adds support for xterm-style mouse reporting driven by touch, so terminal programs that ask for the mouse can respond to taps and scrolling (e.g. `tmux`, `herdr`, `nano`, `vim`, `less`). With this you can tap to move the cursor or click a pane, and drag to scroll.

It's off by default. There's a new "Mouse reporting" switch at the bottom of the terminal emulation settings; when it's off, touch behaves exactly as it did before.

Once mouse reporting is enabled *and* the remote program has actually requested it (via the DECSET sequences the app watches for), touch maps to mouse events like this:

- **Tap** → a left-button press and release at the cell you touched (a click).
- **Vertical drag** → mouse wheel events, one per row of movement, at the cell under your finger. Dragging down scrolls up, matching how scrolling feels everywhere else on the phone.
- **Long-press and hold still** → this is the escape hatch for selecting text. Holding without moving hands the gesture back to the terminal's own text selection (magnifier, handles, drag-to-extend, the works), and no mouse events are sent for the rest of that gesture. So you can still copy text out of a mouse-aware program.
- **Horizontal drag** → left alone, so swiping between sessions still works.
- **Two fingers** → left alone, so pinch-to-zoom still works.

When mouse reporting isn't active, none of this pointer handling is installed at all, so there's zero change to existing touch behaviour.

### DIFF
--- a/app/src/main/java/org/connectbot/ui/screens/console/ConsoleScreen.kt
+++ b/app/src/main/java/org/connectbot/ui/screens/console/ConsoleScreen.kt
@@ -131,6 +131,7 @@ import org.connectbot.service.AuthBanner
 import org.connectbot.service.DisconnectReason
 import org.connectbot.service.PromptRequest
 import org.connectbot.service.TerminalBridge
+import org.connectbot.terminal.MouseTracking
 import org.connectbot.terminal.ProgressState
 import org.connectbot.terminal.SelectionController
 import org.connectbot.terminal.Terminal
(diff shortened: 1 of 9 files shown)


tony5:
I am looking forward to this!

kruton:
Why did you make a parallel terminal emulator for this? It is easily hooked up in termlib.

paulburton:
Now depends on https://github.com/connectbot/termlib/pull/253 going in first.

PR #2322, opened 2026-07-21
Reduce network queries during startup

### LINKED ISSUE #2296: Crash on Android 16: Binder transaction failure (ENOSPC) on startup due to excessive Binder calls in ConnectivityMonitor.init()
## Bug description

ConnectBot crashes immediately on launch with `DeadSystemException` caused by Binder buffer exhaustion. The app's UI is briefly displayed but then the process is killed.

```
E/IPCThreadState(12345): Binder transaction failure. id: ..., BR_*: 29201, error: -28 (No space left on device)
E/JavaBinder(12345): !!! FAILED BINDER TRANSACTION !!!  (parcel size = 96)
E/AndroidRuntime(12345): FATAL EXCEPTION: main
E/AndroidRuntime(12345): Process: org.connectbot, PID: 12345
E/AndroidRuntime(12345): DeadSystemException: The system died; earlier logs will point to the root cause
```

## Root cause

Commit `33f85fa7` ("Better network connectivity tracking", part of v1.10.6) changed `ConnectivityMonitor.init()` to perform significantly more synchronous Binder transactions on the main thread during app startup.

**Before (v1.10.5 — no crash):**
```kotlin
fun init() {
    connectivityManager.registerNetworkCallback(request, networkCallback)
    val activeNetwork = connectivityManager.activeNetwork
    if (activeNetwork != null) {
        updateNetworkInfo(activeNetwork)
    }
}
```
→ ~3 Binder transactions in init()

**After (v1.10.6 — crash):**
```kotlin
fun init() {
    connectivityManager.registerNetworkCallback(request, networkCallback)
    connectivityManager.registerDefaultNetworkCallback(defaultNetworkCallback)
    for (network in connectivityManager.allNetworks) {
        updateNetworkInfo(network)
    }
}
```
→ ~6+ Binder transactions in init()

Combined with all other startup Binder calls (`enableEdgeToEdge()`, `bindService()`, `WindowInsetsController`, theme SideEffect, etc.), the app makes **~17-20 small Binder transactions within <1 second** on process start. On some Android 16 devices with reduced per-process Binder buffer limits, this exhausts the buffer and kills the process with `-ENOSPC`.

## Steps to reproduce

1. Install ConnectBot v1.10.6 or later on a device running Android 16
2. Launch the app
3. App crashes with `DeadSystemException`

## Expected behavior

App should start without crashing.

## Environment

- Android version: 16 (API 37)
- Device: Vivo device (Android 16 build)
- ConnectBot version: v1.10.6+

## Suggested fix

**Option 1 (recommended, minimal risk):**  
Defer the `allNetworks` enumeration loop to the next main looper iteration so it doesn't block the critical startup path:

```kotlin
fun init() {
    connectivityManager.registerNetworkCallback(request, networkCallback)
    connectivityManager.registerDefaultNetworkCallback(defaultNetworkCallback)

    Handler(Looper.getMainLooper()).post {
        for (network in connectivityManager.allNetworks) {
            updateNetworkInfo(network)
        }
    }
}
```

**Option 2 (more conservative):**  
Revert to using `connectivityManager.activeNetwork` for the initial sync (single Binder call) and let the registered callbacks populate the remaining networks asynchronously:

```kotlin
fun init() {
    connectivityManager.registerNetworkCallback(request, networkCallback)
    connectivityManager.registerDefaultNetworkCallback(defaultNetworkCallback)

    val activeNetwork = connectivityManager.activeNetwork
    if (activeNetwork != null) {
        updateNetworkInfo(activeNetwork)
    }
}
```

Relates to commit: `33f85fa79c6a72d8e06a988ff163a120dc19c855`, file: `app/src/main/java/org/connectbot/service/ConnectivityMonitor.kt`

nindanaoto:
Fixes #2296 using Option 2.
- initialize connectivity state from only the active network instead of enumerating every network during startup
- let the registered network callback populate other matching networks asynchronously
- add a Robolectric regression test verifying startup queries activeNetwork and does not use allNetworks

This reduces synchronous Binder transactions during startup and removes use of the deprecated allNetworks API from initialization.

### DIFF
--- a/app/src/main/java/org/connectbot/service/ConnectivityMonitor.kt
+++ b/app/src/main/java/org/connectbot/service/ConnectivityMonitor.kt
@@ -1,6 +1,6 @@
 /*
  * ConnectBot: simple, powerful, open-source SSH client for Android
- * Copyright 2025 Kenny Root
+ * Copyright 2025-2026 Kenny Root
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -130,9 +130,11 @@ class ConnectivityMonitor(
         connectivityManager.registerNetworkCallback(request, networkCallback)
         connectivityManager.registerDefaultNetworkCallback(defaultNetworkCallback)
 
-        // Initialize all active networks
-        for (network in connectivityManager.allNetworks) {
-            updateNetworkInfo(network)
+        // Initialize the active network. Other matching networks will be
+        // populated by networkCallback.
+        val activeNetwork = connectivityManager.activeNetwork
+        if (activeNetwork != null) {
+            updateNetworkInfo(activeNetwork)
         }
     }
 
--- a/app/src/test/kotlin/org/connectbot/service/ConnectivityMonitorTest.kt
+++ b/app/src/test/kotlin/org/connectbot/service/ConnectivityMonitorTest.kt
@@ -26,6 +26,7 @@ import org.junit.Test
 import org.junit.runner.RunWith
 import org.mockito.Mock
 import org.mockito.Mockito.mock
+import org.mockito.Mockito.never
 import org.mockito.Mockito.verify
 import org.mockito.Mockito.`when`
 import org.mockito.MockitoAnnotations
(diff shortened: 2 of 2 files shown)


kruton (approved):

nindanaoto:
Fixed conflict.

Issue #729, opened 2019-12-06
Line feed after post-login command not working / Windows 10


thanoskat:
Although the return character after the command works on a linux server as expected (i.e. command executes), on a windows server it copies the command in prompt and ignores all return characters.

 - Device: Xiaomi Mi A2 Lite
 - OS: Android 9
 - ConnectBot Version: 1.9.6
 - Server OS: Windows 10
 - Server version: 7.7.2.1 (windows built in openssh server)

kruton:
I assume this is a multi-line post-login command? (i.e., the line return is integrated into the post-login command)

thanoskat:
Uh it is a one-liner followed by a return character

keighrim:
I'm having this issue too. My guess is cmd.exe interprets lf alone not as an enter. I couldn't find an easy way to insert cr with android keyboards, so I wasn't able to test my hypothesis.

fejjj:
Logging on to say, this is still an issue.

keighrim:
I was able to work around this limitation by using this app, [UnicodePad](https://play.google.com/store/apps/details?id=jp.ddo.hotmist.unicodepad&hl=en_US&gl=US), to create a `cr-lf` sequence and copy-paste it into the connectbot cmd field.

GoldVinyl:
> I was able to work around this limitation by using this app, [UnicodePad](https://play.google.com/store/apps/details?id=jp.ddo.hotmist.unicodepad&hl=en_US&gl=US), to create a `cr-lf` sequence and copy-paste it into the connectbot cmd field.

Can confirm this works, just make sure you use the unicode symbol U+000D and paste it right after the post-login command, and it works fine.

Issue #75, opened 2015-04-15
Running commands-on-connect via local shell only sometimes works


kruton:
It looks like there is a race condition where we send the commands to be run to the shell before it's actually ready.

![15 - 1 1](https://cloud.githubusercontent.com/assets/50407/7165939/9affe2ea-e35f-11e4-9f98-3824e73eccc2.png)

raineth:
This occurs for me on the official OTA build of 4.4 on my Oneplus One:

```Build: KTU84Q
Build fingerprint: 'oneplus/bacon/A0001:4.4.4/KTU84Q/XNPH05Q:user/release-keys'
Kernel: Linux version 3.4.0-cyanogenmod-gc73a4ec (build04@cyanogenmod) (gcc version 4.7 (GCC) ) #1 SMP PREEMPT Wed Jan 28 11:41:35 PST 2015```

This problem also occurred back in mid-February, before the recent string of ConnectBot releases.  

The problem appears to be that cyanogenmod runs the busybox "resize" command from /system/etc/mkshrc.  The resize command determines the terminal size by sending+receiving terminal control sequences and then feeds the discovered size to termios via the TIOCSWINSZ ioctl.  ConnectBot usually manages to send the post-login automation commands while/before 'resize' is still reading back the cursor position, causing the command(s) to be discarded.  A hacky solution to this problem (and there may not be a better one) would be to sleep briefly before sending the commands.

PR #2354, opened 2026-09-02
Fix post-login automation being sent too early and never entered

pdecat:
Post-login automation has two independent bugs that combine to make it look flaky. One commit each.

## 1. Sent before anything can read it

`TerminalBridge.onConnected()` called `injectString(host.postLogin)` the instant `SSH.finishConnection()` had sent the shell request. `startShell()` only sends that request, it doesn't wait for the remote to spawn anything, so the bytes go out before there is a shell to read them.

Whatever is on the far end at that moment eats them. The clearest case is Tailscale SSH in check mode: tailscaled accepts the session, prints its "visit this URL to authenticate" text, and only spawns the PTY once the browser approval lands. Anything sent beforehand is consumed by the check handler or lost when the real PTY appears, so the command is never echoed and never runs. When the check is still cached no interaction happens, the shell starts immediately, and the same code works. Hence "only sometimes".

Worth ruling out, since it looks like the obvious culprit: this is **not** a missing flush. `TerminalBridge.requestFlush()` has no callers, but sshlib's `ChannelOutputStream.write()` sends a channel-data packet immediately, and its `flush()` is a no-op that only checks the closed flag. The bytes really do leave the device; the remote discards them.

So hold the commands until the remote looks ready:

- A new per-host **Wait for text** field names something to watch for in the remote output. Plain substring against the raw stream, escape sequences included, so a coloured prompt matches on its visible tail. Capped at 120 s, which comfortably outlasts a browser round trip.
- Left empty, the commands go out once the remote has sent something and then stayed quiet for 500 ms. Capped at 10 s.
- Both caps fall back to sending anyway, so a trigger that never matches costs latency rather than a silently skipped command. Existing hosts keep working with no configuration.

The waiter is armed before the relay starts, otherwise a local shell's first prompt can beat it.

This should also cover #75, the same race on the local shell transport, though I haven't reproduced that one.

## 2. Typed but never entered

`injectString()` sends the field verbatim, with no line terminator, so the last command lands on the prompt without running. That was invisible while the string was being swallowed; once it reliably arrives, it just sits there typed.

The field also carries plain line feeds, since that is what the text box produces. A Unix tty takes those as line terminators, but Windows OpenSSH does not, which is why post-login has never worked there even for people who added newlines by hand (#1128, #729).

So send what the Enter key sends: normalise every line ending to a carriage return, and make sure the text ends in one. The remote tty turns it back into the newline the shell reads, and Windows gets the character it was waiting for.

This happens in `schedulePostLogin()` rather than `injectString()`, because that same call pastes the clipboard, and a paste must never enter itself.

## Schema

Adds `post_login_wait_for` to `hosts`. Room 8 to 9 via AutoMigration, schema JSON committed.

## Testing

13 new unit tests cover the waiter and the line-ending conversion: idle settle and its restart on further output, both timeouts, substring match across chunk boundaries, match through escape sequences on a coloured prompt, and the LF / CRLF / CR cases. The coroutine tests use virtual time, so they don't sleep.

`./gradlew :app:testOssDebugUnitTest :app:testGoogleDebugUnitTest :app:lintOssDebug` is green.

`./gradlew check test` fails for me in the `lintAnalyze*UnitTest` / `lintAnalyze*AndroidTest` tasks, but that looks flaky and unrelated to this change: the same command fails the same way on an unmodified `main` here, and re-running it on an unchanged tree flips between pass and fail. It is a Kotlin FIR crash inside lint ("this is a bug in lint or one of the libraries it depends on") while analysing `AppStateScreensTest.kt` and `HiltTestRunner.kt`, neither of which this PR touches. Mentioning it in case it is already known.

Also ran on a device against a Tailscale SSH host that requires the browser check, which is where I hit this: post-login now runs reliably after the round trip.

### DIFF
--- a/app/schemas/org.connectbot.data.ConnectBotDatabase/9.json
+++ b/app/schemas/org.connectbot.data.ConnectBotDatabase/9.json
@@ -1,0 +1,675 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 9,
+    "identityHash": "f27f2bf92b5ef129e66e9f45ecfcec8e",
+    "entities": [
+      {
+        "tableName": "hosts",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, `nickname` TEXT NOT NULL, `protocol` TEXT NOT NULL, `username` TEXT NOT NULL, `hostname` TEXT NOT NULL, `port` INTEGER NOT NULL, `host_key_algo` TEXT, `last_connect` INTEGER NOT NULL, `color` TEXT, `use_keys` INTEGER NOT NULL, `use_auth_agent` TEXT, `post_login` TEXT, `post_login_wait_for` TEXT, `pubkey_id` INTEGER NOT NULL, `want_session` INTEGER NOT NULL, `compression` INTEGER NOT NULL, `stay_connected` INTEGER NOT NULL, `quick_disconnect` INTEGER NOT NULL, `scrollback_lines` INTEGER NOT NULL, `use_ctrl_alt_as_meta_key` INTEGER NOT NULL, `jump_host_id` INTEGER, `profile_id` INTEGER, `ip_version` TEXT NOT NULL DEFAULT 'IPV4_AND_IPV6')",
+        "fields": [
+          {
+            "fieldPath": "id",
+            "columnName": "id",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "nickname",
+            "columnName": "nickname",
+            "affinity": "TEXT",
+            "notNull": true
+          },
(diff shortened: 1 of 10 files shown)


kruton:
I'm not a big fan of this approach. It's very narrowly scoped to one problem.

By the way, you can just type a carriage return in the post-login automation if you need to send it. That's why it's multi-line. If we always send CR after the post-login automation someone is going to complain their far side goes into a TUI that needs specific keypresses or something.

pdecat:
> By the way, you can just type a carriage return in the post-login automation if you need to send it

That's what I was doing before, and it does not always work, there's a race.

kruton:
I meant the second part of your change.

I'll make something more flexible that will satisfy this use case as well.

kruton:
I merged #2417 which should satisfy your usecase. It will be in the next beta.

pdecat:
Nice, thanks !

Issue #1672, opened 2025-12-05
Bell sounds even when ConnectBot is in background


kruton:
## Bug description
The terminal bell is sounding even when ConnectBot is in the background and the screen is off.

## Steps to reproduce
1. Connect to a host
2. Type something like: `sleep 10; echo -e "\a"`
3. Turn off / lock screen before the sleep expires 

## Expected behavior
Either no bell or a notification posted

PR #2355, opened 2026-09-02
Notify on bell only while the UI is not visible

Mathtician:
## Problem

With **Background notifications** enabled, a terminal bell while ConnectBot is in the background (or the screen is off) vibrates instead of posting a notification. The notification only appears if the activity has already been destroyed.

Cause: since the Compose rewrite, `MainActivity` binds `TerminalManager` in `onCreate` and unbinds in `onDestroy`, so the `isUiBound` flag from #1682 stays true whenever the activity exists. Two consequences:

- The bell event still reaches the (alive) `ConsoleViewModel`, which plays the bell in the background; `sendActivityNotification` is suppressed by the gate. This is the other half of #1672.
- A bell from a host other than the one being shown does nothing: the ViewModel asks for a notification, but the same gate suppresses it.

## Change

- `TerminalManager` tracks `isUiVisible` through `ProcessLifecycleOwner` (new dependency `androidx.lifecycle:lifecycle-process`, same version as the other lifecycle artifacts). The bind/unbind flag is removed.
- `TerminalBridge.onBell` decides at the source, as the pre-Compose code did with `parent.isShown()`: not visible → `sendActivityNotification`; visible → emit the bell event and let `ConsoleViewModel` choose between playing the bell (host being shown) and a notification (another host).
- `sendActivityNotification` now only checks the preference; callers decide when a notification is appropriate.

| Situation | Before | After |
|---|---|---|
| App hidden / screen off | bell sound or vibration | notification |
| Visible, bell from the shown host | bell | bell |
| Visible, bell from another host | nothing | notification |

Known remaining gap, unchanged by this PR: a bell while the host list is shown (console ViewModel not alive) produces nothing.

## Testing

- `./gradlew spotlessCheck :app:lintOssDebug :app:testOssDebugUnitTest` pass.
- Verified on a Samsung S24 (Android 16) with two SSH hosts: `sleep 10; printf '\a'` then backgrounding or locking the phone posts the notification with no vibration; staying on the console vibrates with no notification; switching to the other host's session posts the notification.

Related: #1672, #1682.


kruton:
Thanks for the contribution. This was annoying me as well! 😄

PR #2392, opened 2026-09-21
Offer to save host when user edits host

### LINKED ISSUE #2329: Can't edit connection port
### Bug description

After a host connection is saved (ssh type), if you later edit the port then the change is not taken into account even if port is updated in the UI.

### Steps to reproduce

- Add a host connectionwith port 555.
- Edit the connection to change the port (my conn was on 4022) and save host
- Connect to edited connection,
- The connection port is still 555

### Expected behavior

Port should be updated to 4022 when connecting after connection is edited.

### Screenshots

_No response_

### ConnectBot version

1.10.9.53

### Android device

samsunggalaxy zflip5

### Android version

android 16

### Installed from

GitHub releases

### Server OS

_No response_

### SSH software and version

_No response_

### Key type used

_No response_

### Additional context

_No response_

nindanaoto:
Fixes #2329

Saving a host edit previously launched persistence asynchronously and immediately navigated away from the editor. The editor could therefore be destroyed before Room persisted the update.

This affected all fields saved by the host editor, not only the SSH port—for example hostname, username, nickname, protocol, authentication settings, profiles, jump hosts, IP version, and terminal options.

Make host saving suspend until persistence completes, and navigate back only after it returns.

Tests:
- `./gradlew :app:testOssDebugUnitTest --tests org.connectbot.ui.screens.hosteditor.HostEditorViewModelTest`


kruton:
I think we should have a different color FAB with a checkmark come up from the bottom when the user makes a change that is saveable. That's probably more intuitive and will announce itself to accessibility users.

nindanaoto:
The current version shows like this.

nindanaoto:
Fixed CI.

PR #2406, opened 2026-09-22
Add inline images preference to profiles

kruton:
termlib now supports Kitty and iTerm2 style inline images. Add a preference in Profiles (defaulting to Ask) for users to enjoy this new feature.


kruton:
@nindanaoto sorry you'll have to rebase database changes around this.

